PR #17, opened 2025-08-15
chore: Add CI

### LINKED ISSUE #7: Add CI
https://github.com/apache/arrow/blob/main/.github/workflows/csharp.yml

kou:
## What's Changed

This is based on https://github.com/apache/arrow/blob/main/.github/workflows/csharp.yml but packaging job isn't included. Let's work on it as a separated task.

This copies https://github.com/apache/arrow/tree/main/format to `format/`.

This copies scripts:
* https://github.com/apache/arrow/blob/apache-arrow-21.0.0/ci/scripts/csharp_build.sh
* https://github.com/apache/arrow/blob/apache-arrow-21.0.0/ci/scripts/csharp_test.sh

Closes #7.

### DIFF
--- a/format/README.md
+++ b/format/README.md
@@ -1,0 +1,25 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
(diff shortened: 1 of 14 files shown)


kou:
CI result on my fork: https://github.com/kou/arrow-dotnet/actions/runs/16986434875/job/48156245498

format/README.md
```diff
@@ -0,0 +1,25 @@
...
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Arrow Protocol Files
+
+This folder contains binary protocol definitions for the Arrow columnar format
+and other parts of the project, like the Flight RPC framework.
+
+This is a copy of https://github.com/apache/arrow/tree/main/format .
```
adamreeve, line 25: Would it make sense for these format files to live in a new `arrow-format` repository, so they can be included as a submodule and kept up to date in one place rather than need to be manually updated?

I briefly looked at other Arrow implementation repositories (go, swift, js, rust) and as far as I can tell only the Rust implementation also has a copy of these files. I guess this is because the other implementations have checked in the generated code, or they use hand-written implementations rather than generated code.

These files probably won't change very often though so I think it's fine to store copies here and manually update files as needed. And if we did want to change this, it should be a follow-up change and not block this PR.

adamreeve (approved):
:+1: looks good to me

PR #47, opened 2025-09-12
feat: Bounds checking for Flatbuf should be enabled in the default build

### LINKED ISSUE #48: ArrowStreamReader can crash due to invalid field name lengths

`ArrowStreamReader` can read past the end of an allocated buffer within an unsafe context when reading the schema from a bad Arrow IPC stream. Specifically, the unsafe version of [ByteBuffer.GetStringUTF8](https://github.com/apache/arrow-dotnet/blob/710f6e47bd20c5a0dca471a27d67c2fff65d581c/src/Apache.Arrow/Flatbuf/FlatBuffers/ByteBuffer.cs#L694) doesn't validate that its parameters are within the buffer's bounds. When reading the name of a field from the schema message, the length of the field name is read from the buffer in [Table.__string](https://github.com/apache/arrow-dotnet/blob/710f6e47bd20c5a0dca471a27d67c2fff65d581c/src/Apache.Arrow/Flatbuf/FlatBuffers/Table.cs#L66). If the stream contains a length that exceeds the size of the buffer, the reader will read past the end.

In my own testing, I've been able to crash the process with an `AccessViolationException` if the out of bounds read is large enough, or have it populate the field name with data from surrounding memory if the read isn't too large.

The following reproduction case creates a valid IPC stream that only has a schema with a single column. It then edits bytes 108-111, which is the length of the column's name (originally 5) to something larger. Setting it to 50 prints a bunch of junk to the console (from the memory after the buffer), while setting it to 500 causes an access violation.

```csharp
using System;
using System.IO;
using Apache.Arrow;
using Apache.Arrow.Ipc;
using Apache.Arrow.Types;

const int FieldNameLengthOffset = 108;
const int FakeFieldNameLength = 50; // Changing this to something larger like 500 should crash

using MemoryStream stream = new();
Schema schema = new(
    [new Field("index", Int32Type.Default, nullable: false)],
    metadata: []);
using (ArrowStreamWriter writer = new(stream, schema, leaveOpen: true))
{
    writer.WriteStart();
}

stream.Position = FieldNameLengthOffset;
using (StreamWriter writer = new(stream, leaveOpen: true))
{
    writer.Write(FakeFieldNameLength);
}
stream.Position = 0;
using ArrowStreamReader reader = new(stream);
var name = reader.Schema[0].Name;
Console.WriteLine($"Field name: {name}");
```

CurtHagenlocher:
## What's Changed

Enables bounds checking for Flatbuf as input can't generally be trusted.
Adds a test for a malformed column name length.
Makes some fixes required to be able to benchmark the change and demonstrate that there's no significant regression.

Closes #48.

### DIFF
--- a/src/Apache.Arrow/Apache.Arrow.csproj
+++ b/src/Apache.Arrow/Apache.Arrow.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <DefineConstants>$(DefineConstants);UNSAFE_BYTEBUFFER;BYTEBUFFER_NO_BOUNDS_CHECK;ENABLE_SPAN_T</DefineConstants>
+    <DefineConstants>$(DefineConstants);UNSAFE_BYTEBUFFER;ENABLE_SPAN_T</DefineConstants>
     
     <Description>Apache Arrow is a cross-language development platform for in-memory data. It specifies a standardized language-independent columnar memory format for flat and hierarchical data, organized for efficient analytic operations on modern hardware.</Description>
   </PropertyGroup>
--- a/src/Apache.Arrow/Flatbuf/FlatBuffers/ByteBuffer.cs
+++ b/src/Apache.Arrow/Flatbuf/FlatBuffers/ByteBuffer.cs
@@ -693,6 +693,7 @@ namespace Google.FlatBuffers
 #if ENABLE_SPAN_T && UNSAFE_BYTEBUFFER
         public unsafe string GetStringUTF8(int startPos, int len)
         {
+            AssertOffsetAndLength(startPos, len);
             fixed (byte* buffer = &MemoryMarshal.GetReference(_buffer.ReadOnlySpan.Slice(startPos)))
             {
                 return Encoding.UTF8.GetString(buffer, len);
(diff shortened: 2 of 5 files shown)


src/Apache.Arrow/Apache.Arrow.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <DefineConstants>$(DefineConstants);UNSAFE_BYTEBUFFER;BYTEBUFFER_NO_BOUNDS_CHECK;ENABLE_SPAN_T</DefineConstants>
+    <DefineConstants>$(DefineConstants);UNSAFE_BYTEBUFFER;ENABLE_SPAN_T</DefineConstants>
```
adamreeve, line 5: Just checking we're happy with leaving `UNSAFE_BYTEBUFFER` enabled? As far as I understand, bounds checks should prevent any invalid reads and this should be safe in theory, but it means bugs in the Flatbuffers implementation (like missing bounds checks) could still cause memory safety issues. It sounds like this can increase performance significantly though so is worth keeping enabled (https://flatbuffers.dev/languages/c_sharp/#conditional-compilation-symbols).

adamreeve (approved):
Looks good to me thanks Curt.

src/Apache.Arrow/Apache.Arrow.csproj
CurtHagenlocher: I didn't try benchmarking without `UNSAFE_BYTEBUFFER` but I did audit all the code and confirmed that there was only one missing bounds check.

PR #73, opened 2025-09-16
Remove unnecessary allocation in ArrowStreamWriter

### LINKED ISSUE #41: [C#] Improve memory allocation of ArrowStreamWriter cctor

`ArrowStreamWriter` allocates non trivial amount of byte[][] and byte[] objects. In our scenario where we use new ArrowStreamWriter for each Recordbatch it creates significant memory overhead. 

| Method                 | Mean      | Error     | StdDev     | Ratio | RatioSD | Gen0   | Gen1   | Allocated 
|----------------------- |----------:|----------:|-----------:|------:|--------:|-------:|-------:|----------:|
| ArrowStreamWriterCctor | 843.42 ns | 37.066 ns | 108.124 ns |  1.17 |    0.24 | 0.7687 | 0.0248 |    **9656 B** | 

The majority of allocated data is comming from [instantiating ArrayPool](https://github.com/apache/arrow/blob/main/csharp/src/Apache.Arrow/Ipc/ArrowStreamWriter.cs#L666)

The `ArrayPool<byte>.Create()` is very expensive operation as it preallocate the whole pool :

https://github.com/dotnet/runtime/blob/b6127f9c7f6bab00186ec43d4a332053a1d02325/src/libraries/System.Private.CoreLib/src/System/Buffers/ConfigurableArrayPool.cs#L43-L46

| Method                 | Mean      | Error     | StdDev     | Ratio | RatioSD | Gen0   | Gen1   | Allocated |
|----------------------- |----------:|----------:|-----------:|------:|--------:|-------:|-------:|----------:
| ArrayPoolCreate        | 739.13 ns | 40.097 ns | 117.597 ns |  1.03 |    0.23 | 0.6428 | 0.0172 |    **8072 B** |

There is several options how to solve this unnecessary allocation:

### Limit the pool size
This is the simplest change we can do but not perfect. As the pool is used only for arrays of size 4 or 8, we don't need to cfreate full `ArrayPool`, but limit it by parameters:  `ArrayPool<byte>.Create(8, 50)`
Here is comparison of full and limited pool:

| Method                 | Mean      | Error     | StdDev     | Ratio | RatioSD | Gen0   | Gen1   | Allocated | Alloc Ratio |
|----------------------- |----------:|----------:|-----------:|------:|--------:|-------:|-------:|----------:|------------:|
| ArrayPoolCreate        | 739.13 ns | 40.097 ns | 117.597 ns |  1.03 |    0.23 | 0.6428 | 0.0172 |    8072 B |        1.00 |
| ArrayPoolCreateSmall   |  68.02 ns |  3.041 ns |   8.823 ns |  0.09 |    0.02 | 0.0414 |      - |     520 B |        0.06 |

### Use ArrayPool<byte>.Shared
 
Actually there is no reason to create our own array pool because the `Stream.Write(ReadOnlySpan<byte> buffer)` becaue it rent from ArrayPool<byte>.Shared anyway in the default behaviour: https://github.com/dotnet/runtime/blob/b6127f9c7f6bab00186ec43d4a332053a1d02325/src/libraries/System.Private.CoreLib/src/System/IO/Stream.cs#L912-L924

### Don't use Array pool at all

As the array pool is used only in `WriteIpcMessageLengthAsync` and `WriteIpcMessageLengthAsync` it should be enought to have shared array used as a buffer. This would imply the ArrowStreamWriter is not thread safe.

### Component(s)

C#

aik-jahoda:
## What's Changed

ArrowStreamWriter allocates 8k of memory by creating a new array pool. This change introduces a shared buffer instead of pool to reduce the allocations. 

The array pool is used to rent small arrays (8 bytes) but the pool allocates much bigger arrays (8kb in total)
The Array pool has access time overhead for small arrays compared to direct allocation.

Results from benchmarks:

Old implementation:
| Method     | BatchLength | ColumnSetCount | Mean       | Error     | StdDev    | Allocated |
|----------- |------------ |--------------- |-----------:|----------:|----------:|----------:|
| WriteBatch | 10000       | 10             |   6.118 ms | 0.1215 ms | 0.3345 ms | 248.53 KB |
| WriteBatch | 10000       | 14             |   9.788 ms | 0.1910 ms | 0.3396 ms | 324.12 KB |
| WriteBatch | 300000      | 10             | 119.351 ms | 3.1713 ms | 9.3008 ms | 248.53 KB |
| WriteBatch | 300000      | 14             | 136.697 ms | 2.9229 ms | 8.4799 ms | 324.12 KB |

New implementation:
| Method     | BatchLength | ColumnSetCount | Mean       | Error     | StdDev     | Median     | Allocated |
|----------- |------------ |--------------- |-----------:|----------:|-----------:|-----------:|----------:|
| WriteBatch | 10000       | 10             |   5.925 ms | 0.2057 ms |  0.6001 ms |   5.843 ms | 240.64 KB |
| WriteBatch | 10000       | 14             |   8.908 ms | 0.2743 ms |  0.8002 ms |   8.778 ms | 316.23 KB |
| WriteBatch | 300000      | 10             |  94.835 ms | 1.7872 ms |  3.7699 ms |  93.892 ms | 240.64 KB |
| WriteBatch | 300000      | 14             | 147.995 ms | 3.6873 ms | 10.6975 ms | 144.591 ms | 316.23 KB |

Closes #41.

### DIFF
--- a/src/Apache.Arrow/Ipc/ArrowFileWriter.cs
+++ b/src/Apache.Arrow/Ipc/ArrowFileWriter.cs
@@ -14,6 +14,7 @@
 // limitations under the License.
 
 using System;
+using System.Buffers;
 using System.Buffers.Binary;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -221,7 +222,7 @@ namespace Apache.Arrow.Ipc
 
             // Write footer length
 
-            using (Buffers.RentReturn(4, out Memory<byte> buffer))
+            using (ArrayPool<byte>.Shared.RentReturn(4, out Memory<byte> buffer))
             {
                 int footerLength;
                 checked
@@ -292,7 +293,7 @@ namespace Apache.Arrow.Ipc
 
             cancellationToken.ThrowIfCancellationRequested();
 
-            using (Buffers.RentReturn(4, out Memory<byte> buffer))
+            using (ArrayPool<byte>.Shared.RentReturn(4, out Memory<byte> buffer))
             {
                 int footerLength;
                 checked
--- a/src/Apache.Arrow/Ipc/ArrowStreamWriter.cs
+++ b/src/Apache.Arrow/Ipc/ArrowStreamWriter.cs
@@ -611,8 +611,6 @@ namespace Apache.Arrow.Ipc
 
         protected Stream BaseStream { get; }
 
-        protected ArrayPool<byte> Buffers { get; }
-
         private protected FlatBufferBuilder Builder { get; }
 
         protected bool HasWrittenSchema { get; set; }
@@ -663,7 +661,6 @@ namespace Apache.Arrow.Ipc
             _leaveOpen = leaveOpen;
             _allocator = allocator ?? MemoryAllocator.Default.Value;
 
-            Buffers = ArrayPool<byte>.Create();
             Builder = new FlatBufferBuilder(1024);
             HasWrittenSchema = false;
 
@@ -1277,7 +1274,7 @@ namespace Apache.Arrow.Ipc
 
         private void WriteIpcMessageLength(int length)
         {
-            using (Buffers.RentReturn(_options.SizeOfIpcLength, out Memory<byte> buffer))
+            using (ArrayPool<byte>.Shared.RentReturn(_options.SizeOfIpcLength, out Memory<byte> buffer))
             {
                 Memory<byte> currentBufferPosition = buffer;
                 if (!_options.WriteLegacyIpcFormat)
@@ -1294,7 +1291,7 @@ namespace Apache.Arrow.Ipc
 
         private async ValueTask WriteIpcMessageLengthAsync(int length, CancellationToken cancellationToken)
         {
-            using (Buffers.RentReturn(_options.SizeOfIpcLength, out Memory<byte> buffer))
+            using (ArrayPool<byte>.Shared.RentReturn(_options.SizeOfIpcLength, out Memory<byte> buffer))
             {
                 Memory<byte> currentBufferPosition = buffer;
                 if (!_options.WriteLegacyIpcFormat)


aik-jahoda:
I like the idea of making the buffer explicitly for endianity operation. Unfortunately, the stream.WriteAsync doesn't have an overload with `ReadOnlySpan<byte>`  so it will disallow us to use stack alloc . And I believe the async overload will be preferred by users. As we are writing a small amount of data, I tend to think about implementing the `WriteIpcMessageLengthAsync` over sync  `IntBuffer.WriteLittleEndian` so we can leverage zero allocation.

It would be good to support this by benchmark, but it would heavily depend on stream implementation.

aik-jahoda:
I will try to explore solution without buffer, by using https://learn.microsoft.com/en-us/dotnet/api/system.buffers.binary.binaryprimitives.reverseendianness?view=net-9.0#system-buffers-binary-binaryprimitives-reverseendianness(system-int32)

CurtHagenlocher:
> Unfortunately, the stream.WriteAsync doesn't have an overload with `ReadOnlySpan<byte>` so it will disallow us to use stack alloc.

Right; of course, because the allocation has to live until the async call finishes.

> I will try to explore solution without buffer, by using https://learn.microsoft.com/en-us/dotnet/api/system.buffers.binary.binaryprimitives.reverseendianness?view=net-9.0#system-buffers-binary-binaryprimitives-reverseendianness(system-int32)

I suspect the buffer is unavoidable in the async case, but let us know what you decide.

aik-jahoda:
I added your proposal, with async over sync. If you think we should have it fully async, I can add the async variant into the helper.

aik-jahoda:
I have pushed the change with `ArrayPool<byte>.Shared`

If we decide to do a stackalloc for sync method, I propose to do it in separate PR and discuss the solution in a new issue to keep this PR simple.

CurtHagenlocher (approved):

PR #100, opened 2025-10-03
Add serialization helpers for Schema and RecordBatch

### LINKED ISSUE #31: Expose lower-level APIs to consume partial Arrow data

The wire protocols for BigQuery and Databricks expose data formatted as Arrow IPC, but break up the schema and the data rows such that the existing C# libraries can't read the data directly. Instead, the ADBC drivers for these data sources have to construct new structures in memory for the sole purpose of deserialization.

If there were lower-level APIs in the C# Arrow library, then these could be used instead for a gain in efficiency.

CurtHagenlocher:
Adds helpers for Schema and RecordBatch to allow them to be serialized and deserialized individually. Refactors Flight-related code slightly to make use of the new helpers.

Closes #31.

### DIFF
--- a/src/Apache.Arrow/C/CArrowArrayExporter.cs
+++ b/src/Apache.Arrow/C/CArrowArrayExporter.cs
@@ -184,6 +184,7 @@ namespace Apache.Arrow.C
 
             cArray->n_buffers = 1;
             cArray->buffers = (byte**)sharedOwner.Allocate(IntPtr.Size);
+            cArray->buffers[0] = null;
 
             cArray->n_children = batch.ColumnCount;
             cArray->children = null;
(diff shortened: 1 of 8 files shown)


src/Apache.Arrow/C/CArrowArrayExporter.cs
```diff
@@ -184,6 +184,7 @@ private unsafe static void ConvertRecordBatch(ExportedAllocationOwner sharedOwne
 
             cArray->n_buffers = 1;
             cArray->buffers = (byte**)sharedOwner.Allocate(IntPtr.Size);
+            cArray->buffers[0] = null;
```
CurtHagenlocher, line 187: This was already checked-in as a separate change. Will rebase if necessary.

Issue #111, opened 2025-10-06
Allow export of managed buffers via C API


CurtHagenlocher:
### Describe the enhancement requested

You can't export buffers backed by `ReadOnlyMemory` instead of `IMemoryOwner`.
You can't continue to access the array from C# after exporting it.

If it's not possible to implement these things "safely", then at least there should be an opt-in way to do them "unsafely".

PR #112, opened 2025-10-06
Allow C Data Interface export of ReadOnlyMemory-backed buffers

CurtHagenlocher:
This PR partly addresses #111 by allowing C Data Interface export of buffers backed by `ReadOnlyMemory` rather than an `IMemoryOwner`. This is useful if you want to export data that has been read from an IPC file or stream for example. Previously you would need to first copy all buffers for this to work.

Because this is an experimental feature and potentially risky, it must currently be enabled by setting the global static `CArrowArrayExporter.EnableManagedMemoryExport` to `true`.

This change was originally submitted by @adamreeve as [https://github.com/apache/arrow/pull/40992](https://github.com/apache/arrow/pull/40992). My only contribution is disabling the feature by default.

### DIFF
--- a/src/Apache.Arrow/ArrowBuffer.cs
+++ b/src/Apache.Arrow/ArrowBuffer.cs
@@ -16,6 +16,7 @@
 using System;
 using System.Buffers;
 using System.Runtime.CompilerServices;
+using Apache.Arrow.C;
 using Apache.Arrow.Memory;
 
 namespace Apache.Arrow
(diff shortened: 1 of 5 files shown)


CurtHagenlocher:
@adamreeve, I'd like to enable experimental native compilation of C# ADBC drivers -- which requires this feature or something like it. I'm mitigating the risk of the change by marking it as experimental and effectively making it "opt-in" by consuming code.

adamreeve (approved):
:+1: putting this behind an experimental feature makes sense to me

PR #251, opened 2026-02-01
Use unsigned long integers for computing bit-length of fixed-width types via the CData ABI

### LINKED ISSUE #250: CData's `ImportFixedWidthBuffers` method has signed arithmetic overflow failure during array length calculation when importing a large list with a large fixed width type.

https://github.com/apache/arrow-dotnet/blob/70ab46691df24225e2407ad3cc0d22fd534eb18b/src/Apache.Arrow/C/CArrowArrayImporter.cs#L473

I have a data file which includes very large arrays of 64-bit wide types like `double`. In one case this is `65_651_042` values and the bit width is `64`. Of course, when you multiply them together you get `4_201_666_688` a number too large to fit in a 32-bit signed integer, nearly too big for a 32-bit unsigned integer. After dividing by 8 to convert from bits to bytes again, the value falls within range of a 32-bit integer, but before that can happen, overflow occurs:
```
Unhandled exception: System.AggregateException: One or more errors occurred. (Arithmetic operation resulted in an overflow.)
 ---> System.OverflowException: Arithmetic operation resulted in an overflow.
   at Apache.Arrow.C.CArrowArrayImporter.ImportedArrowArray.ImportFixedWidthBuffers(CArrowArray* cArray, Int32 bitWidth)
   at Apache.Arrow.C.CArrowArrayImporter.ImportedArrowArray.GetAsArrayData(CArrowArray* cArray, IArrowType type)
   at Apache.Arrow.C.CArrowArrayImporter.ImportedArrowArray.ProcessListChildren(CArrowArray* cArray, IArrowType type)
   at Apache.Arrow.C.CArrowArrayImporter.ImportedArrowArray.GetAsArrayData(CArrowArray* cArray, IArrowType type)
   at Apache.Arrow.C.CArrowArrayImporter.ImportedArrowArray.ProcessStructChildren(CArrowArray* cArray, IReadOnlyList`1 fields)
   at Apache.Arrow.C.CArrowArrayImporter.ImportedArrowArray.GetAsArrayData(CArrowArray* cArray, IArrowType type)
   at Apache.Arrow.C.CArrowArrayImporter.ImportedArrowArray.GetAsRecordBatch(Schema schema)
   at Apache.Arrow.C.CArrowArrayImporter.ImportRecordBatch(CArrowArray* ptr, Schema schema)
   at Apache.Arrow.C.CArrowArrayStreamImporter.ImportedArrowArrayStream.ReadNextRecordBatchAsync(CancellationToken cancellationToken)
```

PR incoming.

Platform: Windows 10 (64-bit)
dotnet version: 9
Apache.Arrow version: 22.1.0

mobiusklein:
## What's Changed

Uses unsigned 64-bit integers to compute the intermediate bit length of an array to avoid arithmetic overflow. 

I could factorize this expression differently to avoid the casts, but I didn't think this was a sufficiently heavy code path that it'd need more attention.

Closes #250 .

### DIFF
--- a/src/Apache.Arrow/C/CArrowArrayImporter.cs
+++ b/src/Apache.Arrow/C/CArrowArrayImporter.cs
@@ -470,7 +470,10 @@ namespace Apache.Arrow.C
                 int length = checked((int)cArray->offset + (int)cArray->length);
                 int valuesLength;
                 if (bitWidth >= 8)
+                {
+                    Debug.Assert(bitWidth % 8 == 0);
                     valuesLength = checked((int)((ulong)length * (ulong)bitWidth / 8ul));
+                }
                 else
                     valuesLength = checked((int)BitUtility.RoundUpToMultipleOf8(length) / 8);
 


src/Apache.Arrow/C/CArrowArrayImporter.cs
```diff
@@ -470,7 +470,7 @@ private ArrowBuffer[] ImportFixedWidthBuffers(CArrowArray* cArray, int bitWidth)
                 int length = checked((int)cArray->offset + (int)cArray->length);
                 int valuesLength;
                 if (bitWidth >= 8)
-                    valuesLength = checked(length * bitWidth / 8);
+                    valuesLength = checked((int)((ulong)length * (ulong)bitWidth / 8ul));
```
adamreeve: If we can assume that `bitWidth` is always a multiple of 8 then we could just use `length * (bitWidth / 8)`.

If we can't assume that, then this code is incorrect because we round down when dividing by 8 but should round up.

Eg. if `bitWidth = 9` and `length = 1`, we'd need 2 bytes to store all values, not just 1.

I can see we assume `bitWidth` is a multiple of 8 [here for example](https://github.com/apache/arrow-dotnet/blob/70ab46691df24225e2407ad3cc0d22fd534eb18b/src/Apache.Arrow/Arrays/ArrayDataConcatenator.cs#L387), and I can't think of a situation in which that wouldn't be true except for booleans which fall through to the `bitWidth < 8` case below. Maybe we should check that `bitWidth % 8 == 0` and throw an exception if that isn't true?

adamreeve: I don't mind too much which form of expression is used here, as you say this shouldn't be particularly performance sensitive.

I'm just concerned that there's a hidden assumption that isn't being validated and isn't obvious when first reading this piece of code. That was a problem with the original code but it would be nice to fix that if we're touching it now.

From looking a bit closer, the `bitWidth` doesn't come from the C Data Interface input but from the instance of `FixedWidthType` being used, so we don't need to worry about invalid input and this should always be a multiple of the byte size. So maybe just adding a debug assertion here would help make the code clearer:
```suggestion
                {
                    Debug.Assert(bitWidth % 8 == 0);
                    valuesLength = checked((int)((ulong)length * (ulong)bitWidth / 8ul));
                }
```

Yes we can assume a byte will always be 8 bits, but it is possible to define a `FixedWidthType` instance where `BitWidth % 8 != 0`, although that would be a bit insane and terrible for performance.

PR #257, opened 2026-02-08
feature: add selection, filtering, primitive comparison, formatting and limited casting

### LINKED ISSUE #254: A few starter generic compute kernels

I have been hand-writing a pile of generic kernels for numerical arrays and wanted to know what would be required to contribute them upstream.

So far I have implementations for:
- `public static Array Filter(Array array, BooleanArray mask)`
- `public static BooleanArray Equal<T>(PrimitiveArray<T> lhs, T rhs) where T: struct, INumber<T>`
- `public static BooleanArray Equal<T>(PrimitiveArray<T> lhs, PrimitiveArray<T> rhs) where T: struct, INumber<T>`
- `public static Array Take(Array array, IList<int> indices)`
- `public static Int64Array CastInt64<T>(PrimitiveArray<T> array) where T: struct, INumber<T>`
- `public static Int32Array CastInt32<T>(PrimitiveArray<T> array) where T : struct, INumber<T>`
- `public static FloatArray CastFloat<T>(PrimitiveArray<T> array) where T : struct, INumber<T>`
- `public static DoubleArray CastDouble<T>(PrimitiveArray<T> array) where T : struct, INumber<T>`

I also have a generic visitor for converting `Large*` arrays to their regular counterparts so that they may work with the other machinery in the library. I am less confident that is useful to others, or that it is truly correct.

Is there a particular pattern you would want these to follow or is the notion not suitable for the .NET ecosystem? I am less familiar with .NET as an ecosystem.

Thank you.

mobiusklein:
## What's Changed

I've added an `Apache.Arrow.Operations` module as recommended in #254 to hold some utilities for generically working with primitive arrays. 

Included:
 - Array subset selection by index or boolean mask
 - Primitive array and string array equality and comparison
 - Limited numerical aggregations
 - A recursive, not quite complete pretty-printing routine I wrote for debugging. 
 - A small suite of tests to exercise the fiddly bits of indices

Closes #254 

I would like some early feedback to see if I am going in the right direction. As .NET isn't a language I work in routinely, I am likely missing common names or idioms for these operations. I also do not know the degree to which the JIT is doing loop unrolling to make these homogenous operations more performant, though if it needs to be hand rolled that is a strictly later problem.


src/Apache.Arrow.Operations/Apache.Arrow.Operations.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+    </PropertyGroup>
+
+    <PropertyGroup>
+        <TargetFrameworks>net8.0</TargetFrameworks>
```
CurtHagenlocher: Can we target netstandard2.0 also or do you expect it to be hard to support?

src/Apache.Arrow.Operations/Comparison.cs
```diff
@@ -0,0 +1,390 @@
...
+    /// Negate a boolean array, flipping true to false, false to true. Nulls remain null
+    /// </summary>
+    /// <param name="mask"></param>
+    /// <param name="allocator"></param>
+    /// <returns></returns>
+    public static BooleanArray Invert(BooleanArray mask, MemoryAllocator? allocator = null)
+    {
+        var builder = new BooleanArray.Builder();
+        builder.Reserve(mask.Length);
+        foreach (var val in mask)
```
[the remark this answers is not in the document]
CurtHagenlocher: The same approach could also be used for the binary operations, except that the validity buffers would need to be ANDed.

In general, I think it's best to implement these primitive operations in terms of `ArrowBuffer` and then have the Array-based methods call into those.

src/Apache.Arrow.Operations/Comparison.cs
```diff
@@ -0,0 +1,390 @@
...
+
+    /// <summary>
+    /// Compare each value in `lhs` to a scalar `rhs`, returning boolean mask
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    /// <param name="lhs"></param>
+    /// <param name="rhs"></param>
+    /// <param name="allocator"></param>
+    /// <returns></returns>
+    public static BooleanArray Equal<T>(PrimitiveArray<T> lhs, T rhs, MemoryAllocator? allocator = null) where T : struct, INumber<T>
```
CurtHagenlocher: This doesn't seem to take null values into account.

src/Apache.Arrow.Operations/Comparison.cs
```diff
@@ -0,0 +1,390 @@
...
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    /// <param name="lhs"></param>
+    /// <param name="rhs"></param>
+    /// <param name="allocator"></param>
+    /// <returns></returns>
+    /// <exception cref="InvalidOperationException"></exception>
+    public static BooleanArray Equal<T>(PrimitiveArray<T> lhs, PrimitiveArray<T> rhs, MemoryAllocator? allocator = null) where T : struct, INumber<T>
+    {
+        var cmp = new BooleanArray.Builder();
```
CurtHagenlocher: Nit: move the allocation until after the length check. Also applies to a few other methods.

src/Apache.Arrow.Operations/Comparison.cs
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    /// <summary>
+    /// Compare each value in `lhs` to a scalar `rhs`, returning boolean mask
+    /// </summary>
+    /// <param name="lhs"></param>
+    /// <param name="rhs"></param>
+    /// <param name="allocator"></param>
+    /// <returns></returns>
+    public static BooleanArray Equal(StringArray lhs, string rhs, MemoryAllocator? allocator = null)
```
CurtHagenlocher: Should `rhs` be nullable to allow a comparison to null?

It's worth calling out in the documentation that this implements C# semantics, so `null == null`. These are different from SQL semantics where the value of `null == X` is `null`. I think that's the right thing to do, but it might be surprising to someone expecting more SQL-like behavior.

CurtHagenlocher (commented):
Thanks so much for this work! I've left initial comments on the first two files and will try to return to it within the next day.

src/Apache.Arrow.Operations/Comparison.cs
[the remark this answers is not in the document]
CurtHagenlocher: I think I probably mean `Vector256` and not `Avx2`, at least initially.

src/Apache.Arrow.Operations/Comparison.cs
wldevries: https://arrow.apache.org/docs/cpp/compute.html#comparisons
That is indeed surprising

src/Apache.Arrow.Operations/Comparison.cs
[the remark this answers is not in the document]
wldevries: If you were to go that route you should consider adding a dependency on System.Numerics.Tensors and for this method specifically [TensorPrimivitives.OnesComplement](https://learn.microsoft.com/en-us/dotnet/api/system.numerics.tensors.tensorprimitives.onescomplement):
```csharp
byte[] dest = new byte[mask.Values.Length];
System.Numerics.Tensors.TensorPrimitives.OnesComplement<byte>(mask.Values, dest);

var valueBuffer = new ArrowBuffer(dest);
var validityBuffer = mask.Data.Buffers[0];
var data = new ArrayData(
    BooleanType.Default, mask.Length, mask.NullCount,
    mask.Offset, new[] { validityBuffer, valueBuffer });

return new BooleanArray(data);
```
The Microsoft docs state it's .NET 10+, but the code above compiles targeting .NET 8.0. I'm not sure, but it seems like a bug in the docs site.

src/Apache.Arrow.Operations/Comparison.cs
mobiusklein: Good point. Is it preferable to use SQL semantics? Alternatively, it could be configured with a flag argument, in which case the default behavior would still need to be documented.

src/Apache.Arrow.Operations/Comparison.cs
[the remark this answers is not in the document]
CurtHagenlocher: > Is it that an array assumes it is the unique owner of its value and validity buffer and might mutate them directly?

It is because an array assumes that it's the unique owner of its buffers, yes. So if you share the buffer and the first array is disposed, the second one will be holding an invalid buffer. To fix this, buffers would need something like a reference count. This would also be very helpful when exporting via the C API, but I think I've convinced myself that the functionality can't be added in a backwards-compatible way.

The library doesn't support mutable arrays either, which is also a gap if you wanted to optimize computation over arrays. A processing engine might determine, for instance that the expression `a + b` could just reuse `a`'s storage as the target of the computation and avoid having to allocate a new array -- if `a` is never going to be used again. I'm skeptical that C# has the right semantics to implement this in a purely "safe" way but haven't put much effort into working anything out. (Of course, types like `StringArray` are extremely hard to mutate in place no matter what the capabilities of the implementation language.)

src/Apache.Arrow.Operations/Apache.Arrow.Operations.csproj
wldevries: DataFrame from the Microsoft.Data.Analysis project could serve as an example. They still have a T4 macro for netstandard2.0, but switched to `INumber`-based methods for other target frameworks.

https://github.com/dotnet/machinelearning/blob/main/src/Microsoft.Data.Analysis/Computations/Arithmetic.netstandard.tt

src/Apache.Arrow.Operations/Comparison.cs
```diff
@@ -0,0 +1,710 @@
...
+    public static ArrowBuffer OnesComplement(ArrowBuffer buffer)
+    {
+        var builder = new ArrowBuffer.BitmapBuilder(buffer.Length * 8);
+        var store = builder.Span;
+        int offset = 0;
+        int size = buffer.Span.Length;
+
+        while ((size - offset) >= 8)
+        {
+            if ((size - offset) >= 64)
```
CurtHagenlocher: Would it make more sense to promote this to the top level? That is,
```
while ((size - offset) >= 64) { ... }
if ((size - offset) >= 32) { ... }
if ((size - offset) >= 16) { ... }
if ((size - offset) >= 8) { ... }
if ((size - offset) > 0) { ... }
```
That should remove one conditional jump and shrink the size of the while loop.

test/Apache.Arrow.Operations.Tests/Apache.Arrow.Operations.Tests.csproj
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <IsWindows
+            Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Windows)))' == 'true'">
+            true</IsWindows>
+        <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+        <!-- Required for PythonNET -->
+        <EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>
```
CurtHagenlocher: This test project isn't using Python, so this can be removed.

test/Apache.Arrow.Operations.Tests/Apache.Arrow.Operations.Tests.csproj
```diff
@@ -0,0 +1,41 @@
...
+        <TargetFrameworks>net8.0</TargetFrameworks>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="Microsoft.NET.Test.Sdk" />
+        <PackageReference Include="xunit" />
+        <PackageReference Include="xunit.skippablefact" />
+    </ItemGroup>
+
+    <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
```
CurtHagenlocher: We don't need this ItemGroup because the project only targets net8.0. We can also remove the conditional from the next ItemGroup.

src/Apache.Arrow.Operations/Select.cs
```diff
@@ -0,0 +1,722 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Returns the sum of all values in the array.
+    /// </summary>
+    /// <typeparam name="T">The numeric type of array elements.</typeparam>
+    /// <param name="array">The input array.</param>
+    /// <param name="nullHandling">How to handle null values.</param>
+    /// <returns>The sum of values, or null if the array is empty, all values are null,
```
CurtHagenlocher: As "sum" has a well-defined zero, I'm curious why you chose to return null when there are no non-null values. (I don't know enough to think it's wrong, but it's not what I would have done.)

src/Apache.Arrow.Operations/Select.cs
```diff
@@ -0,0 +1,722 @@
...
+
+    /// <summary>
+    /// Returns the index of the minimum value in the array (first occurrence).
+    /// </summary>
+    /// <typeparam name="T">The numeric type of array elements.</typeparam>
+    /// <param name="array">The input array.</param>
+    /// <param name="nullHandling">How to handle null values.</param>
+    /// <returns>The index of the minimum value, or null if the array is empty, all values are null,
+    /// or nullHandling is Propagate and any null exists.</returns>
+    public static int? ArgMin<T>(PrimitiveArray<T> array, AggregateNullHandling nullHandling = AggregateNullHandling.Skip)
```
CurtHagenlocher: It's unfortunate that I'd need to make two passes over the array if I wanted to know both the min or max value and the index at which it first occurred. Should these be combined into a single function? There's no extra cost to compute both, just a slightly clunkier method signature.

src/Apache.Arrow.Operations/Comparison.cs
```diff
@@ -0,0 +1,710 @@
...
+    {
+        if (lhs.Length != rhs.Length) throw new ArgumentException("Arrays must have the same length");
+        var combined = BitVectorOps.And(lhs.ValueBuffer, rhs.ValueBuffer);
+        var combinedMask = BitVectorOps.And(lhs.NullBitmapBuffer, rhs.NullBitmapBuffer);
+        var nullCount = BitUtility.CountBits(combinedMask.Span);
+        return new BooleanArray(combined, combinedMask, lhs.Length, nullCount, 0);
+    }
+
+    /// <summary>
+    /// Performa a pairwise boolean OR operation.
```
CurtHagenlocher:
```suggestion
    /// Perform a pairwise boolean OR operation.
```

src/Apache.Arrow.Operations/Comparison.cs
```diff
@@ -0,0 +1,710 @@
...
+        for (var i = offset; i < size; i++)
+        {
+            store[i] = (byte)~buffer.Span[i];
+        }
+        return builder.Build();
+    }
+
+    public static ArrowBuffer And(ArrowBuffer buffer, ArrowBuffer buffer2)
+    {
+        var builder = new ArrowBuffer.BitmapBuilder(buffer.Length * 8);
```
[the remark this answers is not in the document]
CurtHagenlocher: There's the added complication when combining validity buffers that one or both may be null. I think for `And` and `Or` there should probably be a shortcut at the beginning which says "if `lhs` is empty then return `rhs`". For `Xor`, if both buffers are empty then it should return empty. If just one buffer is empty then it should probably be an error.

src/Apache.Arrow.Operations/Select.cs
```diff
@@ -0,0 +1,722 @@
...
+    /// <summary>
+    /// Returns a copy of the positions in the array included in the provided start-end spans. All other values in the array will be
+    /// excluded.
+    /// </summary>
+    /// <param name="array">The array to select from</param>
+    /// <param name="spans">The index ranges to select</param>
+    /// <param name="allocator">The memory allocator to build the new array from</param>
+    /// <returns></returns>
+    /// <exception cref="InvalidOperationException"></exception>
+    public static Array Take(Array array, IList<(int, int)> spans, MemoryAllocator? allocator = null)
```
CurtHagenlocher: `Take` is usually used for the slicing operation (as in "skip/take"). Is there any reason this shouldn't just use the same name as `Filter`? And this verges on bikeshedding, but maybe `Select` is a good name for this operation?

This overload that takes an `IList<(int, int)>`, should this be private or is it more than an implementation detail of the other methods? I'm having trouble envisioning a scenario where this would be used.

CurtHagenlocher (changes requested):
Another partial review; sorry for the delay. This time I looked at everything but am not entirely done commenting. I do think we need a more efficient implementation of `Filter`/`Select`

src/Apache.Arrow.Operations/Comparison.cs
mobiusklein: Combining this with your earlier suggestion about conditioning on hardware acceleration, to cascade with fewer branches:
```csharp
        if (Vector512.IsHardwareAccelerated)
        {
            while ((size - offset) >= 64) { }
        }
        if (Vector256.IsHardwareAccelerated)
        {
            while ((size - offset) >= 32) { }
        }
        while ((size - offset) >= 16) { }
        while ((size - offset) >= 8) { }
        for (var i = offset; i < size; i++) { }
```
This way, the hardware acceleration check hits once, then loops until it runs out of sufficient data to fill the entire register, then drops to the next size down. Going much further for faster SIMD is beyond the scope I had in mind.

src/Apache.Arrow.Operations/Comparison.cs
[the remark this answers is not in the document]
mobiusklein: All good points. Are the semantics of the validity buffer being empty the equivalent to saying "all the positions are valid"?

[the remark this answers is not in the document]
mobiusklein: Nevermind, you answer this later.

luisquintanilla:
Thanks for the offer to adapt this material, and for all the work here - the selection, comparison, casting and formatting kernels are a great start on a compute surface for arrow-dotnet.

I've opened #379 with a focused first slice: SIMD-accelerated `Sum`/`Min`/`Max`/`Mean` aggregation kernels (generic `INumber<T>` + `TensorPrimitives` fast path on net8.0+, per-type scalar fallback on netstandard2.0/net462, validity-aware null handling, and BenchmarkDotNet coverage). It's intentionally small so it's easy to review.

A suggestion for reconciling the two: keep #379 as the aggregation slice, and let the rest of your work land as its own focused slices (selection/filter, comparison, casting, formatting) rather than one large change - that seemed to be the main review concern here. Where I build on your API ideas (for example `ArgMin`/`ArgMax`, `IArrowArray` overloads, or an explicit null-handling option), I'll credit you. Does that split work for you?

mobiusklein:
That works for me, thank you @luisquintanilla . I wrote a bunch of these for another project and tried to upstream them all at once. These should be broken up and reworked, although there are some core changes that might still be valuable for bit vector construction.

PR #260, opened 2026-02-13
feat: Add support for Run-End Encoded arrays (REE) in Arrow .NET

JorgeCandeias:
This PR attempts to add support for Run-End Encoded arrays by following established codebase patterns.

Notably:
- New `ArrowTypeId` added.
- New array type `RunEndEncodedArray` added.
- New visitor method to handle the new array type.
- New entry in the IPC serializer field type switch.
- New `RunEndEncodedType` nested type.
- Basic feature tests.

This PR is missing targeted performance benchmarks as I could not see an established structure to add them into. Please let me know if, where and how you would like me to create these. They could be useful to observe performance variance from various data sparsity patterns and decide what to optimise.

Please let me know if this PR is missing anything else.

### DIFF
--- a/src/Apache.Arrow/Arrays/RunEndEncodedArray.cs
+++ b/src/Apache.Arrow/Arrays/RunEndEncodedArray.cs
@@ -16,208 +16,207 @@
 using System;
 using Apache.Arrow.Types;
 
-namespace Apache.Arrow
+namespace Apache.Arrow;
+
+/// <summary>
+/// Represents a run-end encoded array.
+/// A run-end encoded array stores consecutive runs of the same value more efficiently.
+/// It contains two child arrays: run_ends (Int16/Int32/Int64) and values (any type).
+/// The run_ends array stores the cumulative end positions of each run.
+/// </summary>
+public class RunEndEncodedArray : Array
 {
     /// <summary>
-    /// Represents a run-end encoded array.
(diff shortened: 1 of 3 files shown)


src/Apache.Arrow/Arrays/RunEndEncodedArray.cs
```diff
@@ -0,0 +1,222 @@
...
+        if (runEnds.Length == 0)
+        {
+            return 0;
+        }
+
+        // Get the last run end value which represents the logical length
+        switch (runEnds)
+        {
+            case Int16Array int16Array:
+                return int16Array.GetValue(int16Array.Length - 1) ?? 0;
```
adamreeve: I think a null run-end value here should throw an exception rather than return 0.

CurtHagenlocher (changes requested):
I generally agree with Adam's comments, particularly about throwing on unexpected data. Can you also please update this PR with fixes for the merge conflicts?

JorgeCandeias:
@CurtHagenlocher My apologies for the late reply and thank you for reviewing this PR. I got a bit overwhelmed and haven't had time to apply your comments, which make complete sense. Please feel free to take over this work if you wish. Otherwise I can PR this again later into the new release when I have a breather. Cheers.

Issue #182, opened 2025-11-18
Support additional data types


CurtHagenlocher:
### Describe the enhancement requested

The C# implementation still needs to add support for:

* [x] LargeStringView / LargeBinaryView / LargeListView
* [x] Run-end encoded arrays
* [ ] Tensors
* [x] Extension types in general

JorgeCandeias:
I have a candidate for REE in my fork. Would you like me to raise a PR?

PR #268, opened 2026-02-22
Add extension type support and implement GuidArray

CurtHagenlocher:
## What's Changed

Adds generic support for extension types including an extension type registry. This has been done in what should be a fairly compatible fashion. No extension types are enabled by default, and different consumers of Arrow inside the same process can explicitly designate which extension types they want to make use of.

Packages several extensibility points into a single `ArrowContext`: the compression factory, the allocator and the extension registry.

Implements the `GuidArray` extension type.

Adds tests including interop tests with Python for `GuidArray`.

Partially implements #182.

### DIFF
--- a/test/Apache.Arrow.Tests/ExtensionTypeTests.cs
+++ b/test/Apache.Arrow.Tests/ExtensionTypeTests.cs
@@ -19,7 +19,6 @@ using System.IO;
 using Apache.Arrow.Arrays;
 using Apache.Arrow.Ipc;
 using Apache.Arrow.Types;
-using Microsoft.Win32;
 using Xunit;
 
 namespace Apache.Arrow.Tests


test/Apache.Arrow.Tests/ExtensionTypeTests.cs
```diff
@@ -0,0 +1,355 @@
...
+                }
+            }
+            var guidArray = builder.Build();
+
+            var field = new Field("guids", GuidType.Default, true,
+                new Dictionary<string, string>
+                {
+                    ["ARROW:extension:name"] = "arrow.uuid",
+                    ["ARROW:extension:metadata"] = ""
+                });
```
adamreeve: I think this metadata should be removed here, we want to test that it gets automatically added when writing to IPC right?

test/Apache.Arrow.Tests/ExtensionTypeTests.cs
```diff
@@ -0,0 +1,355 @@
...
+            var registry = new ExtensionTypeRegistry();
+            registry.Register(GuidExtensionDefinition.Instance);
+
+            var clone = registry.Clone();
+
+            Assert.True(clone.TryGetDefinition("arrow.uuid", out _));
+
+            // Mutating clone should not affect original
+            var newRegistry = new ExtensionTypeRegistry();
+            Assert.False(newRegistry.TryGetDefinition("arrow.uuid", out _));
```
adamreeve: This isn't testing what it says it is. I think you want to do something like unregister Guid from the clone and check you can still get it from the original?

test/Apache.Arrow.Tests/ExtensionTypeTests.cs
```diff
@@ -0,0 +1,355 @@
...
+            var registry = new ExtensionTypeRegistry();
+            registry.Register(GuidExtensionDefinition.Instance);
+
+            var guidType = new GuidType();
+            var field = new Field("uuid_field", guidType, true,
+                new Dictionary<string, string>
+                {
+                    ["ARROW:extension:name"] = "arrow.uuid",
+                    ["ARROW:extension:metadata"] = ""
+                });
```
adamreeve: We shouldn't be setting the metadata manually here, we want to test it gets added by the export process.

test/Apache.Arrow.Tests/ExtensionTypeTests.cs
```diff
@@ -16,10 +16,10 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 using Apache.Arrow.Arrays;
 using Apache.Arrow.Ipc;
 using Apache.Arrow.Types;
+using Microsoft.Win32;
```
adamreeve: Looks like this snuck in accidentally?

adamreeve (approved):

PR #271, opened 2026-02-23
chore: Pin Python version to 3.13 as PythonNet doesn't yet support 3.14

### LINKED ISSUE #270: CI: Python integration tests are being skipped in CI

Eg. see https://github.com/apache/arrow-dotnet/actions/runs/22278848542/job/64445654891?pr=268#step:6:163

These tests should fail if they are being skipped in CI due to the check here: https://github.com/apache/arrow-dotnet/blob/bc99008f9b4c1207f43b9f137d0ec428b1787af8/test/Apache.Arrow.Tests/CDataInterfacePythonTests.cs#L90

But it looks like if the tests are skipped due to a Python version mismatch, the check for being in CI isn't used, which is probably wrong.

adamreeve:
Closes #270 

It looks like PythonNet supporting 3.14 might take a while, so I think we should pin the Python version to 3.13 to keep the C Data Interface tests working.

### DIFF
--- a/.github/workflows/rc.yaml
+++ b/.github/workflows/rc.yaml
@@ -156,7 +156,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
-          python-version: 3
+          python-version: 3.13  # PythonNet isn't compatible with 3.14 yet (https://github.com/pythonnet/pythonnet/issues/2610)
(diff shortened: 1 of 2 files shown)


CurtHagenlocher (approved):
It looks like no tests were skipped `Passed!  - Failed:     0, Passed:  1596, Skipped:     0, Total:  1596, Duration: 6 s - Apache.Arrow.Tests.dll (net8.0)` so this fixes the problem.

Issue #96, opened 2025-09-30
Decimal encoding performance improvement


aik-jahoda:
### Describe the enhancement requested

The current implementation of `DecimalXArray.Builder` shows significantly worse performance compared to other array builders in arrow-dotnet, primarily due to BigInteger arithmetic overhead.

| Method                   | Mean        | Error     | StdDev    | Gen0   | Gen1   | Gen2   | Allocated |
|------------------------- |------------:|----------:|----------:|-------:|-------:|-------:|----------:|
| Int32ArrayBuilder_Append |    317.7 ns |   8.32 ns |  24.27 ns | 0.0234 |      - |      - |     296 B |
| Int64ArrayBuilder        |    318.1 ns |   9.22 ns |  26.74 ns | 0.0310 |      - |      - |     392 B |
| StringArrayBuilder       |  1,620.8 ns |  32.50 ns |  84.47 ns | 0.1850 | 0.1717 | 0.0916 |    1928 B |
| Decimal32ArrayBuilder    |  3,514.0 ns |  68.52 ns | 135.26 ns | 0.0458 | 0.0420 | 0.0153 |     528 B |
| Decimal64ArrayBuilder    |  4,400.3 ns |  87.73 ns | 208.50 ns | 0.0992 |      - |      - |    1256 B |
| Decimal128ArrayBuilder   |  5,949.8 ns | 118.44 ns | 216.58 ns | 0.2289 | 0.1068 | 0.0534 |    2456 B |
| Decimal256ArrayBuilder   | 10,367.5 ns | 245.93 ns | 713.47 ns | 0.2747 | 0.1526 | 0.1068 |    3184 B |

The results are based on the following benchmark: https://github.com/aik-jahoda/arrow-dotnet/commit/a2cf2a6d0ca65f029fe80c61f0e70e4218178292

###Root cause
The main performance bottleneck comes from reliance on BigInteger arithmetic, which is much slower than using native integer math. 

Proposed Directions

I’d like to open a discussion on which approach would best align with arrow-dotnet’s goals and .NET ecosystem support:

### 1. Pre-allocate `BigInteger` powers of 10
Pros:
- Straightforward to implement  
- Provides performance gains for all decimal variants across all .NET versions  

Cons:
- Only partial improvement, may still lag compared to native integer arithmetic  

### 2. Use native integral types (`Int128` in .net 7+, `Int64` in older)
Pros:
[...]

PR #292, opened 2026-03-22
Improvements to decimal conversion

### LINKED ISSUE #247: Decimal256Array could be a little more forgiving

The following throws an exception even though it could return a precise result:
```
    var array = new Decimal256Array.Builder(new Decimal256Type(76, 38)).Append(2422.85527600000m).Build();
    System.Console.WriteLine(array.GetValue(0));

```
In general, we could also consider allowing a consumer to say "I'm okay with truncating a value whose scale fits in a `SqlDecimal` or `System.Decimal` but for which I'd lose digits of precision."

CurtHagenlocher:
## What's Changed

  - Fixes DecimalUtility.GetDecimal to correctly handle high-scale values (e.g., Decimal256 with precision 76, scale 38) that previously threw OverflowException — addresses issue #247
  - Adds a fast path using Int128 on .NET 7+ that avoids BigInteger allocation for values that fit in 128 bits
  - Replaces the old fractional-part overflow check (which threw) with FractionToDecimal, which gracefully reduces precision to fit within decimal's ~28-digit mantissa

Closes #247.
Partially addresses #96.

### DIFF
--- a/test/Apache.Arrow.Tests/Decimal128ArrayTests.cs
+++ b/test/Apache.Arrow.Tests/Decimal128ArrayTests.cs
@@ -504,6 +504,27 @@ namespace Apache.Arrow.Tests
             }
 
             [Fact]
+            public void TryGetValueReturnsFalse()
+            {
+                var array = new Decimal128Array.Builder(new Decimal128Type(38, 4))
+                    .Append(SqlDecimal.Parse("100000000000000000000000000000000"))
+                    .Build();
+
+                Assert.False(array.TryGetValue(0, out decimal? value));
+            }
+
+            [Fact]
+            public void TryGetValueCanRound()
+            {
+                var array = new Decimal128Array.Builder(new Decimal128Type(38, 8))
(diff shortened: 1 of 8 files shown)


test/Apache.Arrow.Tests/Decimal128ArrayTests.cs
```diff
@@ -459,6 +459,62 @@ public void AppendRangeSqlDecimal()
...
+            {
+                var array = new Decimal128Array.Builder(new Decimal128Type(38, 20))
+                    .Append(-2422.85527600000m)
+                    .Build();
+
+                Assert.Equal(-2422.85527600000m, array.GetValue(0));
+            }
+
+            [Fact]
+            public void TryGetValueReturnsTrue()
```
adamreeve: Can we have a test case where this fails and returns false?

adamreeve (commented):
Looks good thanks Curt, I just have some minor suggestions

adamreeve (approved):

PR #309, opened 2026-04-08
fix: Validate additional type information when concatenating

### LINKED ISSUE #307: Array Concatenator doesn't validate type details

The ArrayDataConcatenator will concatenate a decimal64(10, 2) and a decimal64(10, 3) without adjusting the values, resulting in silent corruption of data. The same may be true for other parameterized types.

CurtHagenlocher:
## What's Changed

When concatenating types, ensure that incompatible types are not being concatenating (which can result in corrupt data).

Closes #307.


adamreeve (approved):
👍 LGTM

PR #318, opened 2026-04-20
chore(src): add checked conversions where needed

CurtHagenlocher:
## What's Changed

Adds checks to some int64->int32 conversions when reading data through IPC.

### DIFF
--- a/src/Apache.Arrow.Flight/Internal/RecordBatchReaderImplementation.cs
+++ b/src/Apache.Arrow.Flight/Internal/RecordBatchReaderImplementation.cs
@@ -135,7 +135,7 @@ namespace Apache.Arrow.Flight.Internal
                 {
                     case MessageHeader.RecordBatch:
                         var body = _flightDataStream.Current.DataBody.Memory;
-                        return CreateArrowObjectFromMessage(message, CreateByteBuffer(body.Slice(0, (int)message.BodyLength)), null);
+                        return CreateArrowObjectFromMessage(message, CreateByteBuffer(body.Slice(0, checked((int)message.BodyLength))), null);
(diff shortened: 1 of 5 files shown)


src/Apache.Arrow.Flight/Internal/RecordBatchReaderImplementation.cs
```diff
@@ -134,8 +135,14 @@ public override async ValueTask<RecordBatch> ReadNextRecordBatchAsync(Cancellati
...
                     case MessageHeader.RecordBatch:
+                        if (message.BodyLength < 0 || message.BodyLength > int.MaxValue)
+                        {
+                            throw new InvalidDataException(
+                                $"Cannot read batch. Message body of {message.BodyLength} bytes is out of range");
+                        }
+
                         var body = _flightDataStream.Current.DataBody.Memory;
-                        return CreateArrowObjectFromMessage(message, CreateByteBuffer(body.Slice(0, (int)message.BodyLength)), null);
+                        return CreateArrowObjectFromMessage(message, CreateByteBuffer(body.Slice(0, checked((int)message.BodyLength))), null);
```
adamreeve, line 145: Nit: The `checked` is redundant due to the check above, although doesn't really hurt

adamreeve (approved):
Looks good to me thanks Curt

src/Apache.Arrow.Flight/Internal/RecordBatchReaderImplementation.cs
CurtHagenlocher: Maybe there's a nanosecond performance hit!

PR #324, opened 2026-04-22
Add transparent encoding readers and TimestampWithOffset extension type

CurtHagenlocher:
## What's Changed

- Adds an extension method `AsDecodedReadOnlyList<T>()` on `IArrowArray` that returns an `IReadOnlyList<T>` view over plain, dictionary-encoded, and run-end encoded arrays with zero overhead for plain arrays.
- Adds `EnumeratePhysicalIndices()` to `RunEndEncodedArray` for O(n+m) sequential enumeration instead of O(n·log m) from repeated binary searches.
- Implement the `arrow.timestamp_with_offset` canonical extension type with `TimestampWithOffsetArray`, supporting `IReadOnlyList<DateTimeOffset?>` and a builder. The offset_minutes field accepts dictionary and REE encodings via `AsDecodedReadOnlyList`.


src/Apache.Arrow/Arrays/TimestampWithOffsetArray.cs
```diff
@@ -0,0 +1,261 @@
...
+
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        /// <summary>
+        /// Builder for <see cref="TimestampWithOffsetArray"/>.
+        /// </summary>
+        public class Builder
+        {
+            private readonly TimestampArray.Builder _timestampBuilder;
+            private readonly Int16Array.Builder _offsetBuilder;
```
CurtHagenlocher, line 202: We'll leave encoding this via a dictionary or REE as a future change.

src/Apache.Arrow/Arrays/TimestampWithOffsetArray.cs
```diff
@@ -0,0 +1,261 @@
...
+    /// Extension definition for the "arrow.timestamp_with_offset" canonical extension type.
+    /// Storage is a struct with fields "timestamp" (Timestamp(unit, "UTC")) and
+    /// "offset_minutes" (Int16). The offset_minutes field may be dictionary-encoded
+    /// or run-end encoded.
+    /// </summary>
+    public class TimestampWithOffsetExtensionDefinition : ExtensionDefinition
+    {
+        public static readonly TimestampWithOffsetExtensionDefinition Instance = new TimestampWithOffsetExtensionDefinition();
+
+        public override string ExtensionName => "arrow.timestamp_with_offset";
```
CurtHagenlocher, line 33: As far as I can tell, this isn't implemented for pyarrow yet and so no round trip test is possible.

src/Apache.Arrow/Arrays/TimestampWithOffsetArray.cs
```diff
@@ -0,0 +1,261 @@
...
+            return utc.Value.ToOffset(offset);
+        }
+
+        public int Count => Length;
+        public DateTimeOffset? this[int index] => GetValue(index);
+
+        public IEnumerator<DateTimeOffset?> GetEnumerator()
+        {
+            for (int i = 0; i < Length; i++)
+                yield return GetValue(i);
```
adamreeve: If  `_offsetMinutes` is run-end encoded, this will call `FindPhysicalIndex` for each index. This could be optimized by using the enumerator of the offsets.

test/Apache.Arrow.Tests/ReadOnlyListAdapterTests.cs
```diff
@@ -0,0 +1,485 @@
...
+            {
+                Assert.Equal(reader[i], enumerated[i]);
+            }
+        }
+
+        // =============================================================
+        // Assertion helpers
+        // =============================================================
+
+        private static void AssertInt32Values(IReadOnlyList<int?> reader)
```
adamreeve, line 449: There's a `SlicedReeEnumerationIsEfficient` test that covers enumeration of a REE slice, but no basic test of enumeration without slicing, and no test coverage of the dictionary `GetEnumerator` as far as I can tell.

It might make sense to have these `Assert*Values` methods also test enumeration.

src/Apache.Arrow/Arrays/RunEndEncodedArray.cs
```diff
@@ -395,5 +396,69 @@ private static int FindNormalizePhysicalEndExclusive(ReadOnlySpan<long> runEnds,
...
+        int physicalIndex = FindPhysicalIndex(0);
+
+        switch (RunEnds)
+        {
+            case Int16Array int16RunEnds:
+                {
+                    long currentRunEnd = int16RunEnds.GetValue(physicalIndex).Value;
+                    for (int logical = 0; logical < length; logical++)
+                    {
+                        while (logical + offset >= currentRunEnd)
```
adamreeve, line 423: Runs must have a length of at least 1 according to the spec, so I think this could be an `if` instead of `while`, but it's probably good to be defensive. It took me a little while to understand what's going on here though so maybe a brief comment to explain the logic would be useful.

adamreeve (approved):
This looks really useful thanks Curt, I've left a few suggestions

adamreeve (approved):

PR #326, opened 2026-04-23
Cleaned up some odds and ends

CurtHagenlocher:
## What's Changed

Cleaned up some odds and ends:
- Marked projects as AOT-compatible
- Moved some code from `Json` to `VariantJson` for better clarity
- Enabled compression tests on Desktop Framework


src/Apache.Arrow.Flight.Sql/Apache.Arrow.Flight.Sql.csproj
```diff
@@ -1,7 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard2.1</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
```
adamreeve, line 3: Did you intend to drop netstandard2.1 support here? Should this be `netstandard2.1;net8.0`?

src/Apache.Arrow.Compression/Apache.Arrow.Compression.csproj
```diff
@@ -1,14 +1,9 @@
...
   <PropertyGroup>
-    <Description>Provides decompression support for the Arrow IPC format</Description>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(IsWindows)'=='true'">
     <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(IsWindows)'!='true'">
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'net8.0'))">true</IsAotCompatible>
```
adamreeve, line 5: Do you want to add net8.0 as a target framework, otherwise this will always be false right?

src/Apache.Arrow.Flight.Sql/Apache.Arrow.Flight.Sql.csproj
CurtHagenlocher: Yes, this was intentional. From what I can tell, there's no good reason to support netstandard2.1. Anything before net8.0 is out of support and anything after it should be just as compatible with net8.0 as it is with netstandard2.1. If you have a reason to think differently, let me know and I can revert the change (either before or after it's checked in ;) )

adamreeve: Ok no you're right, this change makes sense!

PR #327, opened 2026-04-24
fix: handle file reader cancellation and empty nested concat

InCerryGit:
## Summary

This PR fixes two correctness edge cases found while reviewing the IPC reader and array concatenation paths.

### File reader cancellation

`ArrowFileReaderImplementation.ReadRecordBatchAsync(...)` and
`ReadNextRecordBatchAsync(...)` accepted a caller cancellation token, but the
schema/footer loading step still called `ReadSchemaAsync()` without passing that
token.

That meant a canceled async file read could still continue through schema/footer
I/O before cancellation was observed later in the record batch path. This PR
passes the caller token into schema loading so cancellation is honored before
dictionary and record batch reads begin.

### Empty nested array concatenation

`ArrayDataConcatenator` could drop the required child array structure when
concatenating nested list arrays whose parent arrays were all empty. Nested Arrow
arrays still need a valid zero-length child `ArrayData`; otherwise downstream
array construction can fail or produce structurally invalid nested arrays.

This PR preserves a zero-length child for all-empty nested/list inputs.

### Null-only ListView concatenation

For `ListView` and `LargeListView`, null rows can have `size == 0`, and their
offset values should not contribute to the child slice range. The previous logic
used every offset when computing the child bounds, even for zero-size entries.

This PR computes list-view child bounds only from entries with `size > 0`, so
null-only inputs keep an empty child slice instead of deriving a range from null
rows.

## Validation

- `dotnet test test/Apache.Arrow.Tests/Apache.Arrow.Tests.csproj -c Release --filter "FullyQualifiedName~ArrowFileReaderTests.ReadRecordBatchAsync_HonorsPreCanceledTokenDuringSchemaRead|FullyQualifiedName~ArrowFileReaderTests.ReadNextRecordBatchAsync_HonorsPreCanceledTokenDuringSchemaRead|FullyQualifiedName~ArrowArrayConcatenatorTests.TestConcatenateAllEmpty|FullyQualifiedName~ArrowArrayConcatenatorTests.TestConcatenateNullOnly"`
- `dotnet build Apache.Arrow.sln -c Release`

### DIFF
--- a/src/Apache.Arrow/Arrays/ArrayDataConcatenator.cs
+++ b/src/Apache.Arrow/Arrays/ArrayDataConcatenator.cs
@@ -147,6 +147,17 @@ namespace Apache.Arrow
                         }
                     }
 
+                    if (!hasNonEmptyValue)
+                    {
+                        for (int i = 0; i < offsets.Length; ++i)
+                        {
+                            offsetsBuilder.Append(baseOffset);
+                        }
+
+                        children.Add(child.Slice(0, 0));
+                        continue;
+                    }
+
                     foreach (long offset in offsets)
                     {
                         offsetsBuilder.Append(baseOffset + offset - minOffset);
@@ -202,6 +213,17 @@ namespace Apache.Arrow
                             maxEnd = Math.Max(maxEnd, offsets[i] + sizes[i]);
                             hasNonEmptyValue = true;
                         }
+                    }
+
+                    if (!hasNonEmptyValue)
+                    {
+                        for (int i = 0; i < offsets.Length; ++i)
+                        {
+                            offsetsBuilder.Append(baseOffset);
+                        }
+
+                        children.Add(child.Slice(0, 0));
+                        continue;
                     }
 
                     foreach (int offset in offsets)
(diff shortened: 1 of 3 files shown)


src/Apache.Arrow/Arrays/ArrayDataConcatenator.cs
```diff
@@ -133,13 +133,18 @@ public void Visit(LargeListViewType type)
                     var child = arrayData.Children[0];
                     ReadOnlySpan<long> offsets = arrayData.Buffers[1].Span.CastTo<long>().Slice(arrayData.Offset, arrayData.Length);
                     ReadOnlySpan<long> sizes = arrayData.Buffers[2].Span.CastTo<long>().Slice(arrayData.Offset, arrayData.Length);
-                    var minOffset = offsets[0];
+                    long minOffset = 0;
```
CurtHagenlocher, line 136: Does this work if we're handed a slice of the array?

InCerryGit: Good catch. For slices that contain at least one non-empty value this works because minOffset is initialized from the first size > 0 entry, not from zero.

But for a sliced ListView/LargeListView where every selected entry has size == 0,minOffset stays zero and we may preserve non-zero offsets while producing a zero-length child. I'll update this path to handle the no-non-empty-values case explicitly and normalize those offsets to the current base offset.

CurtHagenlocher: The concatenator is under-tested, for sure; if we had existing coverage for concatenating sliced arrays, it could have caught that.

InCerryGit: Agreed. I added targeted regression coverage for concatenating sliced ListView and LargeListView arrays, including null-only slices, valid empty-list slices, and the case where zero-size slices are concatenated after a non-empty input so the expected normalized offset is a non-zero base offset. 
That should cover the sliced-array gap that allowed this edge case to slip through.

CurtHagenlocher (approved):
Thanks for these fixes!

PR #329, opened 2026-04-25
Refactors VariantValueWriter

CurtHagenlocher:
## What's Changed

Refactors `VariantValueWriter` to use a new `Buffer<>` type to manage the output buffers instead of using `MemoryStream`.

### DIFF
--- a/src/Apache.Arrow.Scalars/Variant/Buffer.cs
+++ b/src/Apache.Arrow.Scalars/Variant/Buffer.cs
@@ -52,6 +52,9 @@ namespace Apache.Arrow.Scalars.Variant
     /// <see cref="Acquire"/> must be called before any write; writes on a
     /// default-initialized buffer throw <see cref="NullReferenceException"/>.
     /// Failing to <see cref="Release"/> leaks the backing array to the GC.
+    /// For the same reason, the local pool must also be cleaned up when done.
+    /// This can be done with <see cref="DrainPool"/>.
+    /// 
     /// </para>
     /// </remarks>
     internal struct Buffer<T>
@@ -92,6 +95,19 @@ namespace Apache.Arrow.Scalars.Variant
                 pool.Push(_buf);
                 _buf = null;
                 _length = 0;
+            }
+        }
+
+        /// <summary>
+        /// Returns every array stashed in <paramref name="pool"/> to
+        /// <see cref="ArrayPool{T}.Shared"/>. Use at end-of-life of the owner
+        /// to release the per-owner cache built up by <see cref="Release"/>.
+        /// </summary>
+        public static void DrainPool(Stack<T[]> pool)
+        {
+            while (pool.Count > 0)
+            {
+                ArrayPool<T>.Shared.Return(pool.Pop());
             }
         }
 
--- a/src/Apache.Arrow.Scalars/Variant/Buffer.cs
+++ b/src/Apache.Arrow.Scalars/Variant/Buffer.cs
@@ -1,0 +1,213 @@
+// Licensed to the Apache Software Foundation (ASF) under one or more
+// contributor license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright ownership.
+// The ASF licenses this file to You under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance with
+// the License.  You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


src/Apache.Arrow.Scalars/Variant/VariantValueWriter.cs
```diff
@@ -314,160 +411,232 @@ public void WriteDecimal16(SqlDecimal value)
...
-            ms.WriteByte(VariantEncodingHelper.MakePrimitiveHeader(VariantPrimitiveType.TimestampNtzNanos));
-            WriteInt64LE(ms, nanos);
+            ref Buffer<byte> buf = ref BeforeWriteValue();
+            buf.Append(VariantEncodingHelper.MakePrimitiveHeader(VariantPrimitiveType.TimestampNtzNanos));
+            buf.WriteInt64LE(nanos);
+        }
+
+        // ---------------------------------------------------------------
+        // Transcode from a VariantReader
+        // ---------------------------------------------------------------
```
CurtHagenlocher: Oops, crossed the streams

src/Apache.Arrow.Scalars/Variant/VariantValueWriter.cs
```diff
@@ -314,160 +411,232 @@ public void WriteDecimal16(SqlDecimal value)
...
+            value.TryWriteBytes(raw, bigEndian: true, out _);
 #else
-            byte[] native = value.ToByteArray();
             // Convert from .NET mixed-endian to big-endian (RFC 4122).
-            _scratch[0] = native[3]; _scratch[1] = native[2]; _scratch[2] = native[1]; _scratch[3] = native[0];
-            _scratch[4] = native[5]; _scratch[5] = native[4];
-            _scratch[6] = native[7]; _scratch[7] = native[6];
-            Buffer.BlockCopy(native, 8, _scratch, 8, 8);
-            ms.Write(_scratch, 0, 16);
+            byte[] native = value.ToByteArray();
```
CurtHagenlocher: Hmm... we have an unsafe helper for this somewhere

CurtHagenlocher: Not accessible from here, so copying the code for now.

src/Apache.Arrow.Scalars/Variant/Buffer.cs
```diff
@@ -0,0 +1,213 @@
...
+{
+    /// <summary>
+    /// Mutable, non-thread-safe growable buffer designed for single-owner use as
+    /// a field on a class (or as a local owned by the same method that created
+    /// it). The backing array is rented from a caller-supplied
+    /// <see cref="Stack{T}"/> so that capacity is "sticky" across reuse cycles
+    /// within one owner — unlike <see cref="ArrayPool{T}.Shared"/>, which
+    /// buckets by size class and may hand back a different array than the one
+    /// last returned.
+    /// </summary>
```
adamreeve, line 31: Might be worth adding a comment that arrays may be added to the pool from `ArrayPool<T>.Shared` by `Acquire` and the caller is responsible for returning all arrays in the Stack back to `ArrayPool<T>.Shared`.

adamreeve (approved):
Looks good to me thanks Curt.

Is the motivation for this just to increase performance by reducing allocations, and do you have any measurements of the effect of this change on allocations?

I saw you made a [comment](https://github.com/apache/arrow-dotnet/pull/328#issuecomment-4317755507) about wanting to refactor `VariantValueWriter` before adding shredding support, but I'm not sure how this helps with shredding.

CurtHagenlocher:
> I saw you made a [comment](https://github.com/apache/arrow-dotnet/pull/328#issuecomment-4317755507) about wanting to refactor `VariantValueWriter` before adding shredding support, but I'm not sure how this helps with shredding.

There is shredding-related code which needs to write a `ReadOnlySpan<byte>` and because `VariantValueWriter` used a `MemoryStream` the span needed to be materialized into an array first. This got me to change how `VariantValueWriter` works, and in doing so I realized that the existing pooled arrays it used were very susceptible to leaking. I have a bias towards fixing already-checked-in code before adding new code, so I submitted this PR first.

src/Apache.Arrow.Scalars/Variant/Buffer.cs
CurtHagenlocher: Thanks! This got me to realize that the arrays being rented in `Buffer` but returned in `VariantValueWriter` was unsatisfying if not actively dangerous. I've moved the drainage code to `Buffer` and added a comment.

adamreeve: 👍 that's much nicer having the array pool interactions all within `Buffer`

PR #331, opened 2026-04-26
perf: improve StringArray builder append paths

InCerryGit:
## Summary

- Avoid temporary byte-array allocations for small `StringArray.Builder.Append(string)` values by encoding into stack memory before appending.
- Pre-reserve offsets, validity, and value-buffer capacity for known-count `AppendRange` inputs.
- Add focused correctness coverage for nulls, empty strings, custom encodings, large-string fallback, collection inputs, and non-collection enumerables.

`AppendRange(ICollection<string>)` now performs a counting prepass to reserve value-buffer capacity before appending, so collection inputs are enumerated twice by design.

## Benchmark

BenchmarkDotNet ShortRun, `StringBuilderAppendBenchmark`, 10,000 ASCII strings of length 32:

| Method | Before | After |
| --- | ---: | ---: |
| `AppendSmallStrings` | 432.0 us / 1.66 MB | 341.0 us / 1157.5 KB |
| `AppendRangeSmallStrings` | 426.2 us / 1.66 MB | 311.8 us / 353.68 KB |

## Validation

- `dotnet test test/Apache.Arrow.Tests/Apache.Arrow.Tests.csproj -c Release --filter "FullyQualifiedName~Apache.Arrow.Tests.StringArrayTests"`
- `rtk dotnet build "Apache.Arrow.sln" -c Release`
- LSP diagnostics clean on changed files
- Code review completed before commit; no blockers found

### DIFF
--- a/src/Apache.Arrow/Arrays/StringArray.cs
+++ b/src/Apache.Arrow/Arrays/StringArray.cs
@@ -30,8 +30,6 @@ namespace Apache.Arrow
 
         public new class Builder : BuilderBase<StringArray, Builder>
         {
-            private const int StackallocByteThreshold = 256;
-
             public Builder() : base(StringType.Default) { }
 
             protected override StringArray Build(ArrayData data)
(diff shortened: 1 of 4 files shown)


src/Apache.Arrow/Arrays/StringArray.cs
```diff
@@ -43,13 +45,54 @@ public Builder Append(string value, Encoding encoding = null)
                 {
                     return AppendNull();
                 }
+
                 encoding = encoding ?? DefaultEncoding;
-                byte[] span = encoding.GetBytes(value);
-                return Append(span.AsSpan());
+
+                int byteCount = encoding.GetByteCount(value);
```
CurtHagenlocher, line 49: Instead of using `stackalloc`, why not just grow the underlying buffer by `byteCount` and `GetBytes` straight into the destination buffer?

CurtHagenlocher: This might require a new function on `BinaryArray.Builder` similar to `IBufferWriter<T>.GetSpan(Int32)`.

InCerryGit: Updated to encode directly into the reserved ValueBuffer span instead of using stackalloc, then resized after encoding succeeds. Thanks.

CurtHagenlocher (approved):
Thanks!

test/Apache.Arrow.Tests/StringArrayTests.cs
```diff
@@ -13,6 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System.Text;
+using System.Collections.Generic;
 using Xunit;
```
CurtHagenlocher, line 18: It looks like the linter wants these sorted alphabetically.

InCerryGit: Done.

PR #33, opened 2025-09-05
chore: Prepare https://arrow.apache.org/dotnet/

kou:
## What's Changed

Add support for publishing to https://arrow.apache.org/dotnet/ : 
* The `main` branch -> https://arrow.apache.org/dotnet/main/
* The `vX.Y.Z` tag -> https://arrow.apache.org/dotnet/X.Y.Z/
* The latest `vX.Y.Z` tag -> https://arrow.apache.org/dotnet/current/
* https://arrow.apache.org/dotnet/ -> https://arrow.apache.org/dotnet/current/

Add support for previewing on fork: https://${GITHUB_ID}.github.io/arrow-dotnet/
Example: https://kou.github.io/arrow-dotnet/

See docs/README.md how to prepare fork to preview.

Add support for commenting preview URL on PR.

Closes #26.

PR #332, opened 2026-04-26
Add Parquet variant shredding support

CurtHagenlocher:
## What's Changed

Implements the Parquet variant shredding spec end-to-end in a new `Apache.Arrow.Operations.Shredding` namespace, alongside minor changes to the base scalar and array types.

Operations.Shredding reader side:
- `ShreddedVariant` / `ShreddedObject` / `ShreddedArray` ref-struct trio exposing typed columns and residual bytes side-by-side.
- `VariantArrayShreddingExtensions` adds `GetShreddedVariant(i)` and `GetLogicalVariantValue(i)` on `VariantArray`.
- `ShredSchema.FromArrowType` derives a shredding schema from an Arrow typed_value type, rejecting unsupported types (uint32, fixed-size-binary(N≠16)).

Operations.Shredding producer side:
  - `VariantShredder` decomposes a column of `VariantValues` against a `ShredSchema` into shared metadata + per-row `ShredResults`.
  - `ShreddedVariantArrayBuilder` assembles those into a shredded `VariantArray` with a `typed_value` Arrow tree matching the schema.

Apache.Arrow changes:
- `VariantExtensionDefinition` accepts `struct<metadata, value?, typed_value?>` layouts in addition to the plain unshredded form.
- `VariantType` gains `IsShredded` / `HasValueColumn` / `HasTypedValueColumn` / `TypedValueField` properties.
- `VariantArray.GetVariantValue` and `GetVariantReader` throw on shredded columns with a pointer to the `Operations.Shredding` extensions.
- The public `VariantArray(IArrowArray)` constructor now infers the `VariantType` (shredded or not) from the storage shape.
- Operations gains a project reference to Apache.Arrow; Apache.Arrow does not reference Operations.

  Apache.Arrow.Scalars changes:
- `VariantValueWriter.CopyValue(VariantReader source)` transcodes a reader into this writer, re-resolving field IDs against the writer's metadata dictionary. Supports cross-dictionary transcoding and multi-source merge-into-one-dictionary workflows.
- `VariantMetadataBuilder.CollectFieldNames(VariantReader source)` is the two-pass companion that accumulates source field names into the target metadata builder.

Validation:
- Conformance tests run against the Iceberg shredded-variant corpus in `apache/parquet-testing` (`test/parquet-testing/shredded_variant/`). `test/shredded_variant_ipc/regen.py` converts each `case-NNN.parquet` to an Arrow IPC file via `pyarrow`; 137 resulting .arrow files are checked in so CI needs no Python. All 128 valid conformance cases pass; 6 schema-invalid and data-invalid cases are rejected with clear errors; 3 "spec-invalid but permissive" INVALID cases are documented as read-without-throw.
- Additional round-trip, reader-style, and builder tests were implemented


src/Apache.Arrow.Scalars/Variant/VariantValue.cs
```diff
@@ -113,6 +113,24 @@ public static VariantValue FromDecimal8(decimal value) =>
...
+        /// want the smallest decimal type that fits the value.
+        /// </summary>
+        public static VariantValue FromDecimal16(SqlDecimal value)
+        {
+            if (value.Data[3] != 0)
+            {
+                SqlDecimal normalized = SqlDecimal.ConvertToPrecScale(value, 38, value.Scale);
+                return new VariantValue(VariantPrimitiveType.Decimal16, (object)normalized);
+            }
+            return new VariantValue(VariantPrimitiveType.Decimal16, (object)value.Value);
```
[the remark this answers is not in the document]
CurtHagenlocher: I think in a followup change I'm going to always store a Decimal16 as a SqlDecimal and vice versa. The current "convert to decimal if it fits" strategy is unnecessarily complicated. Filed #33 to cover this.

src/Apache.Arrow.Operations/Shredding/ShreddingHelpers.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// </summary>
+        public static ShreddedVariant BuildSlot(
+            ShredSchema slotSchema,
+            ReadOnlySpan<byte> metadata,
+            StructArray elementGroup,
+            int index)
+        {
+            StructType elementGroupType = (StructType)elementGroup.Data.DataType;
+            int valueIdx = elementGroupType.GetFieldIndex("value");
+            int typedIdx = elementGroupType.GetFieldIndex("typed_value");
```
CurtHagenlocher, line 40: These should probably be cached; need to take a second look.

CurtHagenlocher: (Shouldn't affect the public API, so can be done as a followup.)

test/shredded_variant_ipc/regen.py
```diff
@@ -0,0 +1,79 @@
...
+Converts shredded-variant Parquet test cases from
+test/parquet-testing/shredded_variant/*.parquet into Arrow IPC (.arrow) files
+under this directory, so that .NET tests can load them without a Parquet
+reader. The Parquet test corpus comes from apache/parquet-testing.
+
+Requires: pyarrow (tested with 23.0).
+
+Run from the repo root:
+
+    python test/shredded_variant_ipc/regen.py
```
adamreeve, line 26: I tested running this and I get a diff in `case-037.arrow`. Any idea why that would be? The test case is `Variant(metadata=VariantMetadata(dict={}), value=Variant(type=UUID, value=f24f9b64-81fa-49d1-b74e-8c09a6e31c56))`. I tested with a few PyArrow versions (23.0.0, 23.0.1, 24.0.0) on x86_64 Linux.

src/Apache.Arrow.Operations/Shredding/ShreddedArray.cs
```diff
@@ -0,0 +1,151 @@
...
+                    ShreddedVariant slot = ShreddingHelpers.BuildSlot(_schema.ArrayElement, _metadata, elementGroup, i);
+                    elements.Add(slot.ToVariantValue());
+                }
+                return VariantValue.FromArray(elements);
+            }
+
+            // No typed list at this row — decode from the residual.
+            if (_residual == null || _residual.IsNull(_row))
+            {
+                throw new InvalidOperationException(
```
adamreeve: Does this not indicate a null value, so we should return `VariantValue.Null`?

adamreeve (commented):
I started reviewing this but didn't get very far, I'll just leave the couple of comments I have for now

test/shredded_variant_ipc/regen.py
CurtHagenlocher: I'm also getting a difference when running under MacOS vs Windows, so I'll investigate that.

CurtHagenlocher: The difference is in the Arrow metadata. On Windows, `ARROW:extension:metadata` comes before `ARROW:extension:name` and on MacOS (and I guess Linux) it comes afterwards. The Arrow C++ code is apparently using a hash-based container to manage these and so the iteration order is implementation-defined.

The difference in output doesn't have an impact on the test. I suspect that people are more likely to try to regenerate the files on MacOS and Linux than on Windows, so I could change the PR to use that ordering -- I'm pretty neutral about that. I will, however, update `regen.py` with a comment about this possible discrepancy.

src/Apache.Arrow.Operations/Shredding/ShreddedArray.cs
CurtHagenlocher: I think there's some ambiguity in the spec -- or at least enough ambiguity for this to have come up either as an issue or on the mailing list -- between `null` at a Parquet/Arrow level and a variant `null`. That said, I'm pretty sure I made the opposite decision in at least two other places so I'll update this one to be consistent.

test/shredded_variant_ipc/regen.py
CurtHagenlocher: I've decided to replace the Windows-generated version with the Mac-generated version if only to see whether you still get a difference :).

adamreeve: Thanks for looking into this, yes I get no diff now compared to the Mac-generated version.

src/Apache.Arrow/Arrays/VariantArray.cs
```diff
@@ -79,14 +92,46 @@ public class VariantType : ExtensionType
...
+        /// <summary>
+        /// True if the storage layout has a <c>typed_value</c> field (shredded).
+        /// </summary>
+        public bool HasTypedValueColumn { get; }
+
+        /// <summary>
+        /// True if the storage layout includes any shredding (has a <c>typed_value</c>
+        /// column, or lacks a <c>value</c> column indicating full shredding).
+        /// </summary>
+        public bool IsShredded => HasTypedValueColumn || !HasValueColumn;
```
adamreeve: The check for `!HasValueColumn` is redundant, if there's no `value` column then there must be a `typed_value` column. Do we need both `IsShredded` and `HasTypedValueColumn`?

test/Apache.Arrow.Operations.Tests/Shredding/ShreddedVariantArrayBuilderTests.cs
```diff
@@ -0,0 +1,424 @@
...
+        {
+            var values = new List<VariantValue>
+            {
+                VariantValue.FromInt32(42),
+                VariantValue.FromString("hello"),
+            };
+            VariantArray array = ShredAndBuild(values, ShredSchema.Unshredded());
+
+            Assert.False(array.IsShredded);
+            AssertRoundTrip(values, ShredSchema.Unshredded());
```
adamreeve: `AssertRoundTrip` could return the shredded array so we don't need to call `ShredAndBuild` twice here and in `Object_FullyShredded`, and there are a few other tests where it might make sense to make assertions about whether the array is shredded, or check if there are any residual values.

src/Apache.Arrow.Operations/Shredding/ShreddedVariantArrayBuilder.cs
```diff
@@ -0,0 +1,513 @@
...
+            // already handles those, but for simplicity we emit the plain binary form.)
+            BinaryArray.Builder metadataBuilder = new BinaryArray.Builder();
+            for (int i = 0; i < rowCount; i++)
+            {
+                metadataBuilder.Append((ReadOnlySpan<byte>)metadata);
+            }
+            BinaryArray metadataArr = metadataBuilder.Build(allocator);
+
+            // value column: residual bytes (or null).
+            BinaryArray valueArr = BuildBinaryColumn(rows, allocator);
```
adamreeve, line 64: Should we omit the value array if values are fully shredded? Probably fine to add that as an optimisation later though if there's a need for it.

test/Apache.Arrow.Operations.Tests/Shredding/ShreddedVariantReaderTests.cs
```diff
@@ -0,0 +1,511 @@
...
+            Assert.Equal(14.3, slot.GetDouble());
+        }
+
+        [SkippableFact]
+        public void GetDecimal_ReadsShreddedDecimal4()
+        {
+            VariantArray array = LoadCase("case-024"); // Decimal4 = 123456.789 (scale 3?)
+            ShreddedVariant slot = array.GetShreddedVariant(0);
+            decimal d = slot.GetDecimal();
+            Assert.NotEqual(0m, d);
```
adamreeve:
```suggestion
            VariantArray array = LoadCase("case-024"); // Decimal4 (scale 4)
            ShreddedVariant slot = array.GetShreddedVariant(0);
            decimal d = slot.GetDecimal();
            Assert.Equal(12345.6789m, d);
```

adamreeve (approved):
This all looks good to me thanks Curt, only a few minor comments.

src/Apache.Arrow.Operations/Shredding/ShreddedVariantArrayBuilder.cs
CurtHagenlocher: My concern with doing that is a hypothetical scenario where we're shredding a column in a very large table. We get the values as an `IArrowArrayStream` instead of an `IArrowArray` and we run each of the batches through `ShredSchemaInferrer`, leaving us with a `ShredSchema`. Now we take a second pass through the `IArrowArrayStream` and shred the batches, one at a time. Each of the batches will need to conform to the shredded schema and we can't just omit `values` in one of them without knowing whether or not it can be omitted in all of them.

In short, I think this would require a separate knob based on the bigger picture.

PR #340, opened 2026-04-27
perf: speed up MemoryStream IPC stream reads

InCerryGit:
## Summary

This improves `ArrowStreamReader` when reading from `MemoryStream` instances that expose their underlying buffer. The reader now uses the exposed buffer for IPC message/schema metadata reads while preserving the existing reader-owned body-buffer boundary.

The change is intentionally scoped to MemoryStream-backed IPC stream reads:

- public/exposed `MemoryStream` can use the fast path
- non-public `MemoryStream` and partial-read streams continue through the fallback stream-read path
- record batch body data is still copied into allocator-owned memory before array construction
- `ArrowMemoryReader` exact-length continuation-token handling is corrected for complete in-memory buffers

## Benchmark

BenchmarkDotNet ShortRun, `ArrowReaderBenchmark`:

| Scenario | Before | After |
|---|---:|---:|
| `ArrowReaderWithMemoryStream_ManagedMemory`, 100000 rows / 1 column | 21629.3 us | 7707.6 us |
| `ArrowReaderWithMemoryStream_ManagedMemory`, 100000 rows / 5 columns | 91112.3 us | 40137.5 us |

## Validation

- `dotnet test test/Apache.Arrow.Tests/Apache.Arrow.Tests.csproj -c Release --filter "FullyQualifiedName~Apache.Arrow.Tests.ArrowStreamReaderTests"`
- `dotnet test test/Apache.Arrow.Compression.Tests/Apache.Arrow.Compression.Tests.csproj -c Release --filter "FullyQualifiedName~Apache.Arrow.Compression.Tests.ArrowStreamReaderTests"`
- `dotnet build Apache.Arrow.sln -c Release`

### DIFF
--- a/src/Apache.Arrow/Ipc/ArrowMemoryStreamReaderImplementation.cs
+++ b/src/Apache.Arrow/Ipc/ArrowMemoryStreamReaderImplementation.cs
@@ -32,6 +32,7 @@ namespace Apache.Arrow.Ipc
     internal sealed class ArrowMemoryStreamReaderImplementation : ArrowStreamReaderImplementation
     {
         private readonly MemoryStream _stream;
+        private readonly Memory<byte> _streamMemory;
 
         public ArrowMemoryStreamReaderImplementation(
             MemoryStream stream,
@@ -42,6 +43,13 @@ namespace Apache.Arrow.Ipc
             : base(stream, allocator, compressionCodecFactory, leaveOpen, extensionRegistry)
         {
             _stream = stream;
+
+            if (!stream.TryGetBuffer(out ArraySegment<byte> streamBuffer))
+            {
+                throw new InvalidOperationException("Expected MemoryStream to expose its backing buffer.");
+            }
+
+            _streamMemory = streamBuffer.Array.AsMemory(streamBuffer.Offset, streamBuffer.Count);
         }
 
         public override ValueTask<RecordBatch> ReadNextRecordBatchAsync(CancellationToken cancellationToken)
@@ -65,7 +73,7 @@ namespace Apache.Arrow.Ipc
             ReadResult result = default;
             do
             {
-                result = ReadMessageFromExposedMemoryStream();
+                result = ReadMessageFromMemory();
             } while (result.Batch == null && result.MessageLength > 0);
 
             return result.Batch;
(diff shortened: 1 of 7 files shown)


test/Apache.Arrow.Benchmarks/ArrowReaderBenchmark.cs
```diff
@@ -29,16 +29,22 @@ namespace Apache.Arrow.Benchmarks
     [MemoryDiagnoser]
     public class ArrowReaderBenchmark
     {
-        [Params(10_000, 1_000_000)]
+        [Params(10_000, 100_000)]
```
CurtHagenlocher: Why did this need to change? Because there are more columns in the batch?

src/Apache.Arrow/Ipc/ArrowStreamReaderImplementation.cs
```diff
@@ -316,5 +512,6 @@ public ReadResult(int messageLength, RecordBatch batch)
                 Batch = batch;
             }
         }
+
```
CurtHagenlocher: nit: remove added blank line

CurtHagenlocher (changes requested):
It feels like there are now effectively two different implementations of `ArrowMemoryReaderImplementation`: one for `MemoryStream`s and one for every other kind of stream. Given that this is an internal class, would it make more sense to have a separate `class ArrowMemoryStreamReaderImplementation : ArrowMemoryReaderImplementation` that handles the `MemoryStream`-specific flavor?

InCerryGit:
> It feels like there are now effectively two different implementations of `ArrowMemoryReaderImplementation`: one for `MemoryStream`s and one for every other kind of stream. Given that this is an internal class, would it make more sense to have a separate `class ArrowMemoryStreamReaderImplementation : ArrowMemoryReaderImplementation` that handles the `MemoryStream`-specific flavor?

That makes sense. The current change does make ArrowStreamReaderImplementation carry both the general stream path and the exposed MemoryStream path. I’ll move the MemoryStream-specific logic into a separate internal implementation while preserving the existing stream-reader ownership semantics for record batch bodies.

test/Apache.Arrow.Tests/ArrowStreamReaderTests.cs
```diff
@@ -315,163 +609,6 @@ [new Field("index", Int32Type.Default, nullable: false)],
             });
         }
 
-        [Fact]
```
CurtHagenlocher, line 318: I think this is a merge error. I recently added these tests and there's no reason to delete them again :D.

src/Apache.Arrow/Ipc/ArrowMemoryStreamReaderImplementation.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        public ArrowMemoryStreamReaderImplementation(
+            MemoryStream stream,
+            MemoryAllocator allocator,
+            ICompressionCodecFactory compressionCodecFactory,
+            bool leaveOpen,
+            ExtensionTypeRegistry extensionRegistry)
+            : base(stream, allocator, compressionCodecFactory, leaveOpen, extensionRegistry)
+        {
+            _stream = stream;
```
CurtHagenlocher, line 45: Consider extracting the `Memory<byte>` in the constructor and reusing it instead of extracting it each time. Otherwise, every time it's extracted it allocates a new object.

src/Apache.Arrow/Ipc/ArrowMemoryStreamReaderImplementation.cs
```diff
@@ -0,0 +1,212 @@
...
+            if (schemaMessageLength == 0)
+            {
+                return;
+            }
+
+            Memory<byte> schemaBuffer = ReadExposedMemory(schemaMessageLength);
+            _schema = MessageSerializer.GetSchema(ReadMessage<Flatbuf.Schema>(CreateByteBuffer(schemaBuffer)), ref _dictionaryMemo, _extensionRegistry);
+        }
+
+        private ReadResult ReadMessageFromExposedMemoryStream()
```
CurtHagenlocher: Consider making the method names more terse now that all the methods work against a `MemoryStream`

src/Apache.Arrow/Ipc/ArrowStreamReaderImplementation.cs
```diff
@@ -47,11 +47,9 @@ protected override void Dispose(bool disposing)
             }
         }
 
-        public override async ValueTask<RecordBatch> ReadNextRecordBatchAsync(CancellationToken cancellationToken)
+        public override ValueTask<RecordBatch> ReadNextRecordBatchAsync(CancellationToken cancellationToken)
         {
-            // TODO: Loop until a record batch is read.
-            cancellationToken.ThrowIfCancellationRequested();
-            return await ReadRecordBatchAsync(cancellationToken).ConfigureAwait(false);
+            return ReadRecordBatchAsync(cancellationToken);
```
CurtHagenlocher, line 53: Perhaps the `ThrowIfCancellationRequested` should be put back? (I don't know that it's needed, but there doesn't seem to be a good reason to remove it.)

CurtHagenlocher (changes requested):
Thanks again! My only real concern now is that some tests appear to have been inadvertently deleted from `ArrowStreamReaderTests`. These should be put back.

CurtHagenlocher (approved):

PR #350, opened 2026-05-04
Port verify_rc to Powershell to make it easier to run on Windows

CurtHagenlocher:
## What's Changed

Ports verify_rc.sh to Powershell to make it easier to run on Windows.

### DIFF
--- a/dev/release/verify_rc.ps1
+++ b/dev/release/verify_rc.ps1
@@ -147,13 +147,21 @@ function Test-Source-Distribution {
 
     Invoke-Block { dotnet build }
 
-    # Python and PyArrow are required for C Data Interface tests.
+    # Python 3 and PyArrow are required for C Data Interface tests.
     if (-not (Test-Path env:PYTHON)) {
+        # On Windows, Python 3 installs as python.exe (no python3.exe).
+        # On Linux/macOS, python3 is the standard name.
         if (Get-Command python3 -ErrorAction SilentlyContinue) {
             $env:PYTHON = "python3"
-        } else {
+        } elseif (Get-Command python -ErrorAction SilentlyContinue) {
             $env:PYTHON = "python"
+        } else {
+            throw "Python is required but neither python3 nor python was found"
         }
+    }
+    $pyVersion = & $env:PYTHON --version 2>&1
+    if ($pyVersion -notmatch 'Python 3\.') {
+        throw "Python 3 is required but found: $pyVersion"
     }
     Invoke-Block { & $env:PYTHON -m pip install pyarrow find-libpython }
     $env:PYTHONNET_PYDLL = (& $env:PYTHON -m find_libpython).Trim()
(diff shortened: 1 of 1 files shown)


CurtHagenlocher:
@adamreeve, this is largely AI-translated; I have no desire to learn Powershell :D. It would be great if you simply verify that this runs for you on a Windows machine (if you have access to one). You'd need to install gpg. I'm going to try it on one or two additional boxes myself.

dev/release/verify_rc.ps1
```diff
@@ -0,0 +1,275 @@
...
+function Test-Source-Distribution {
+    if ([int]$env:VERIFY_SOURCE -le 0) {
+        return
+    }
+
+    Invoke-Block { dotnet build }
+
+    # Python and PyArrow are required for C Data Interface tests.
+    if (-not (Test-Path env:PYTHON)) {
+        if (Get-Command python3 -ErrorAction SilentlyContinue) {
```
adamreeve, line 154: This didn't work correctly for me. I have `python` on my path and it's the real python, but `python3` resolves to `C:\Users\Adam\AppData\Local\Microsoft\WindowsApps\python3.exe`, which is an "App execution alias". When this is run a few lines down, it returns an error code and outputs:
```
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Apps > Advanced app settings > App execution aliases.
```

I checked and there's also a `C:\Users\Adam\AppData\Local\Microsoft\WindowsApps\python.exe` (without the 3).

I could work around this by setting `$env:PYTHON`, so it's not a huge problem, but maybe there's a way to check whether python3 is actually a proper python executable, eg. by running it with a simple argument like `--version`?

adamreeve (approved):
I'm not a powershell expert but this looks sensible and worked for me, besides one issue with Python resolution. I also tested with `$env:VERIFY_SIGN=0` before installing gpg.

dev/release/verify_rc.ps1
CurtHagenlocher: Thanks for calling this out; I'd forgotten that I ended up having to manually copy `python.exe` on top of `python3.exe` which is silly. I'll fix the script instead.

adamreeve: The latest change doesn't fix the issue for me because `Get-Command python3` succeeds, so `$env:PYTHON` is set to `"python3"`, but then it fails when calling ` $env:PYTHON --version` below. Maybe we could use something like this to replace the calls to `Get-Command`:

```powershell
function Test-Python([string]$pythonCommand) {
	if (-not (Get-Command $pythonCommand -ErrorAction SilentlyContinue)) {
		return $false
	}
	$PSNativeCommandUseErrorActionPreference = $false
	& $pythonCommand --version 2>&1 | Out-Null
	$PSNativeCommandUseErrorActionPreference = $true
	return $LASTEXITCODE -eq 0
}
```

CurtHagenlocher: For some reason I didn't correctly understand your comment and so I checked in the change without addressing it. This is an annoying misfeature in Windows; I forgot about it because I'd already disabled it on the machine where I tested the later changes. I'll submit a fix.

PR #351, opened 2026-05-05
Fix Powershell RC verification script to avoid stubs

CurtHagenlocher:
## What's Changed

Th Powershell RC verification script has been fixed to avoid any paths in `WindowsApps` as these are likely to be stubs that point the Microsoft Store rather than being the programs we want.


adamreeve (approved):
👍 This worked on my machine thanks Curt

PR #358, opened 2026-05-18
docs: Remove Run End Encoding from Not Implemented

cjsha:
## What's Changed

This feature is implemented as of v23.0.0 release (PR #308), 
so it was removed from the  "Not Implemented" section of the docs.


adamreeve (approved):
:+1: thanks @cjsha

PR #360, opened 2026-05-18
Fix integration tests by modifying CompareVariadicArrays

### LINKED ISSUE #359: Integration tests failing with Rust producing and .NET consuming binary views
Both the IPC and C Data Interface integration tests are failing the `binary_view` test with Rust producing and .NET consuming.

Eg. see https://github.com/apache/arrow-dotnet/actions/runs/26013406285/job/76458463982#step:13:14196

There's an arrow-rs issue for this here: https://github.com/apache/arrow-rs/issues/9989

CurtHagenlocher:
## What's Changed

Fix integration tests by modifying `CompareVariadicArrays` so that it only does a direct buffer comparison when `_strictCompare` is enabled.

Fixes #359

### DIFF
--- a/test/Apache.Arrow.Tests/ArrowReaderVerifier.cs
+++ b/test/Apache.Arrow.Tests/ArrowReaderVerifier.cs
@@ -362,7 +362,7 @@ namespace Apache.Arrow.Tests
 
                 CompareValidityBuffer(expectedArray.NullCount, _expectedArray.Length, expectedArray.NullBitmapBuffer, expectedArray.Offset, actualArray.NullBitmapBuffer, actualArray.Offset);
 
-                Assert.True(expectedArray.Views.SequenceEqual(actualArray.Views));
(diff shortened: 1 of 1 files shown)


adamreeve:
Rather than remove this completely, we might want to put it behind an `if (_strictCompare)` check? It looks like `_strictCompare` is set to false for the integration tests: https://github.com/apache/arrow-dotnet/blob/40b9d9aeb17712bb8c9b5618bb672175aef87757/test/Apache.Arrow.IntegrationTest/IntegrationCommand.cs#L90

CurtHagenlocher:
> Rather than remove this completely, we might want to put it behind an `if (_strictCompare)` check? It looks like `_strictCompare` is set to false for the integration tests:

That's a good idea; I'd forgotten about that flag but I think it's pretty much meant for this exact scenario. I'm just waiting to validate that the tests pass with this change since we're already twenty minutes in ...

PR #369, opened 2026-06-14
Update ArrowBuffer.cs

kronic:
## What's Changed

Remove allocation ArrowBuffer.Empty

### DIFF
--- a/src/Apache.Arrow/ArrowBuffer.cs
+++ b/src/Apache.Arrow/ArrowBuffer.cs
@@ -25,7 +25,7 @@ namespace Apache.Arrow
         private readonly SharedMemoryHandle _handle;
         private readonly ReadOnlyMemory<byte> _memory;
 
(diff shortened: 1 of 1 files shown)


src/Apache.Arrow/ArrowBuffer.cs
```diff
@@ -25,7 +25,7 @@ namespace Apache.Arrow
         private readonly SharedMemoryHandle _handle;
         private readonly ReadOnlyMemory<byte> _memory;
 
-        public static ArrowBuffer Empty => new ArrowBuffer(Memory<byte>.Empty);
+        public static ArrowBuffer Empty { get; } new ArrowBuffer(Memory<byte>.Empty);
```
CurtHagenlocher:
```suggestion
        public static ArrowBuffer Empty { get; } = new ArrowBuffer(Memory<byte>.Empty);
```

kronic: Ok

CurtHagenlocher (approved):

Issue #375, opened 2026-06-19
Add compute kernels (Sum/Min/Max/Mean) using System.Numerics.Tensors (TensorPrimitives) and Tensor<T> support


luisquintanilla:
## Summary

Implement a small, well-tested compute/kernels library that uses System.Numerics.Tensors (TensorPrimitives and Tensor<T>/TensorSpan<T>) to provide fast aggregation and elementwise kernels for Arrow PrimitiveArray<T>. This converts the repo's documented TODO ("no compute / kernel abstraction") into a working, multi-targeted, SIMD-accelerated compute layer while remaining safe on older frameworks.

## Goals

- Provide production-quality Sum/Min/Max/Mean aggregation kernels for int32/int64/float/double (generic where net8+ allows).  
- Provide elementwise Add/Subtract/Multiply/Divide kernels (scaffolded) that produce new PrimitiveArray<T> results.  
- Implement a simple Arrow Tensor wrapper that can be constructed from Flatbuf Tensor messages and exposed as TensorSpan<T>/Tensor<T> on net8+ (wrapping existing buffers without copying where safe).  
- Add unit tests and benchmarks.  
- Keep changes additive, behind #if NET8_0_OR_GREATER where necessary, and preserve scalar fallbacks for netstandard2.0/net462.

## Current State

- PrimitiveArray<T> already exposes ReadOnlySpan<T> Values (no marshaling needed) and there is an existing MemoryAllocator to allocate result buffers.  
- System.Numerics.Tensors offers span-based SIMD-accelerated primitives (TensorPrimitives) that match Arrow's memory model.  
- This issue defines exact file additions/edits, code skeletons, tests, and a PR checklist so a junior dev can implement and close it.

## Scope / Acceptance criteria

1. New project: src/Apache.Arrow.Compute added and referenced from solutions/projects where appropriate.  
2. Directory.Packages.props updated to include System.Numerics.Tensors v10.0.9 (or later stable).  
3. Aggregations: Add Sum/Min/Max/Mean for int, long, float, double.  
   - Fast path: when array.NullCount == 0 call TensorPrimitives.* on Values.  
   - Null-aware path: correct scalar fallback implementation that respects validity bitmap and NullCount > 0.  
4. Elementwise: scaffold Add/Subtract/Multiply/Divide APIs producing a new PrimitiveArray<T> value buffer and combined validity bitmap.  
[...]

PR #379, opened 2026-06-22
feat: Add compute aggregation kernels (Sum/Min/Max/Mean)

luisquintanilla:
## What's Changed

Adds a new `Apache.Arrow.Compute` project with SIMD-accelerated Sum/Min/Max/Mean aggregation kernels for `PrimitiveArray<T>`, backed by `System.Numerics.Tensors` (`TensorPrimitives`). When `NullCount == 0` the kernels dispatch to `TensorPrimitives` for one vectorized pass over the contiguous `Values` span; with nulls they fall back to a validity-aware scalar loop. The project is AOT-compatible with scalar fallbacks for older target frameworks, and pulls in `System.Numerics.Tensors` via `Directory.Packages.props`.

Includes an xUnit test project (`Apache.Arrow.Compute.Tests`, 11 tests) covering int32, int64, float, and double, with and without nulls, and the empty-array case. The new projects are added to `Apache.Arrow.sln` and the test to `Apache.Arrow.Tests.slnf`.

This implements the aggregation-kernel portion of #375; the elementwise kernels and `Tensor<T>` wrapper described there remain as follow-ups.

Part of #375.


CurtHagenlocher:
This overlaps a bit with #257, at least in spirit. FYI @mobiusklein.

mobiusklein:
I ran out of bandwidth. The scope of my PR was too large to work through in the time I had available. If @luisquintanilla wants to adapt any of that material, please feel free to. Otherwise, I won't be able to cycle back to it in the next month at least.

luisquintanilla:
Thanks folks. I'll take a look and see how to reconcile both PRs where possible. 

To @CurtHagenlocher point, seems like they're both aiming to add operations to `PrimitiveArray` so looks like there's definitely opportunities.

luisquintanilla:
Pushed an update that gets this PR closer to mergeable (keeping it as a draft for now):

- **Multi-targets** `netstandard2.0;net8.0;net462` to match `Apache.Arrow`. On net8.0+ the kernels stay generic over `INumber<T>` and dispatch to `System.Numerics.Tensors` (`TensorPrimitives`) on the null-free fast path; on netstandard2.0/net462 (where generic math and `TensorPrimitives` aren't available) they fall back to per-type scalar kernels for `Int32Array`/`Int64Array`/`FloatArray`/`DoubleArray`.
- **Null handling** follows LINQ semantics (nulls skipped; `Sum` of empty/all-null is zero; `Min`/`Max`/`Mean` throw `InvalidOperationException` when there are no non-null elements), with a validity-aware loop when nulls are present.
- **Benchmarks** added (`Apache.Arrow.Compute.Benchmarks`, BenchmarkDotNet). On a quick run the SIMD `Sum` is roughly 6-7x a naive managed loop over the same buffer at 1M elements, with zero allocation.
- 11 unit tests, green on all target frameworks.

This is the first slice of #375; a roadmap for the follow-on slices (elementwise kernels, `Tensor<T>` interop, and an extended aggregation surface) is on that issue.

**On the project home / naming.** These are compute kernels in the sense of C++ `arrow::compute` and `pyarrow.compute`, so this PR puts them in a dedicated `Apache.Arrow.Compute` project. I noticed the `Apache.Arrow.Operations` name suggested on #254 is now used by the Arrow Variant binary-format codecs, so a separate `Apache.Arrow.Compute` keeps the numeric/array kernels cleanly apart from those codecs. Happy to align with whatever you'd prefer here, @CurtHagenlocher (including a different home for the kernels in #257).

luisquintanilla:
@CurtHagenlocher this one's ready for review. Thanks!

Apache.Arrow.sln
```diff
@@ -1,3 +1,4 @@
+﻿
```
adamreeve: This extra blank line should be removed.

Apache.Arrow.sln
```diff
@@ -264,10 +275,51 @@ Global
...
+		{C9657321-6DFA-411A-AD42-D51C51D0FAF0}.Release|Any CPU.Build.0 = Release|Any CPU
+		{C9657321-6DFA-411A-AD42-D51C51D0FAF0}.Release|x64.ActiveCfg = Release|Any CPU
+		{C9657321-6DFA-411A-AD42-D51C51D0FAF0}.Release|x64.Build.0 = Release|Any CPU
+		{C9657321-6DFA-411A-AD42-D51C51D0FAF0}.Release|x86.ActiveCfg = Release|Any CPU
+		{C9657321-6DFA-411A-AD42-D51C51D0FAF0}.Release|x86.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
 	EndGlobalSection
+	GlobalSection(NestedProjects) = preSolution
```
adamreeve, line 317: This nests the new projects under "src" and "test" folders, but no other projects are nested like this. Can you either remove the nesting or make other projects also nested for consistency.

I like the nesting and it matches the directory layout so my preference would be to also nest the others.

src/Apache.Arrow.Compute/Aggregations.cs
```diff
@@ -0,0 +1,420 @@
...
+                throw new InvalidOperationException("Sequence contains no non-null elements.");
+            }
+
+            if (array.NullCount == 0)
+            {
+                return TensorPrimitives.Min(values);
+            }
+
+            bool set = false;
+            T min = T.Zero;
```
adamreeve: If you add an `IMinMaxValue<T>` constraint, you could initialize min to `T.MaxValue` and not need `set`. And similarly for Max.

src/Apache.Arrow.Compute/Aggregations.cs
```diff
@@ -0,0 +1,420 @@
...
+
+            T sum = array.Sum();
+            return double.CreateChecked(sum) / count;
+        }
+#else
+        // netstandard2.0 / net462 fallback: generic math and TensorPrimitives are unavailable, so the
+        // kernels are provided as per-type overloads backed by validity-aware scalar loops. The null
+        // semantics match the generic net8.0+ implementation above.
+
+        private const string NoElements = "Sequence contains no non-null elements.";
```
adamreeve: This could be moved above the `#if NET8_0_OR_GREATER` and reused for the .NET 8 implementations too.

adamreeve (commented):
Thanks @luisquintanilla, I've left a few minor comments.

To fix the CI tests, you'll need to update `dev/release/verify_rc.ps1` and `dev/release/verify_rc.sh` to make the new test project reference the nuget packages. See [here](https://github.com/apache/arrow-dotnet/blob/e40a12b7ecf91a82e7b295f7c4bbe58936cb71d8/dev/release/verify_rc.sh#L182-L188) and [here](https://github.com/apache/arrow-dotnet/blob/e40a12b7ecf91a82e7b295f7c4bbe58936cb71d8/dev/release/verify_rc.ps1#L223-L229).

The documentation build is also failing, it's not immediately obvious to me how to fix that.

luisquintanilla:
Thanks for the review @adamreeve 

I believe I've addressed all the issues and feedback.

CurtHagenlocher:
> * **Null handling** follows LINQ semantics (nulls skipped; `Sum` of empty/all-null is zero; `Min`/`Max`/`Mean` throw `InvalidOperationException` when there are no non-null elements), with a validity-aware loop when nulls are present.

I think some of these LINQ semantics would be surprising to most data folks. Is there a difficulty in typing `Min` / `Max` / `Mean` to return a `Nullable<T>` and having them return null when there are no non-null elements? Otherwise, it puts a burden on the caller to either do an expensive pre-scan or to incur the rather steep cost of trapping an exception.

> **On the project home / naming.** These are compute kernels in the sense of C++ `arrow::compute` and `pyarrow.compute`, so this PR puts them in a dedicated `Apache.Arrow.Compute` project. I noticed the `Apache.Arrow.Operations` name suggested on #254 is now used by the Arrow Variant binary-format codecs, so a separate `Apache.Arrow.Compute` keeps the numeric/array kernels cleanly apart from those codecs. Happy to align with whatever you'd prefer here, @CurtHagenlocher (including a different home for the kernels in #257).

I don't have a super strong preference about this, other than to insist that if we have an additional assembly then there must be clear guidelines about how to decide between the two where functionality goes. I don't see a particularly clear line between "compute operations" and "non-compute operations" but I also haven't put very much thought into it.

adamreeve:
> I think some of these LINQ semantics would be surprising to most data folks

FWIW, I checked how PyArrow's compute module behaves and it returns None in these scenarios. Making these return a `Nullable` would also make it more obvious to users that they need to handle this case. It's easier to forget to catch an exception.

luisquintanilla:
Hey folks,

Got caught up in other things so I didn't have a chance to get back to this. 

Thanks for the feedback. I've addressed it in the latest commit. 

## Change summary

### Return `Nullable<T>`

All of the operations now return `Nullable<T>` when there are no non-null elements. This aligns with Arrow's default ([`ScalarAggregateOptions`](https://github.com/apache/arrow/blob/main/cpp/src/arrow/compute/api_aggregate.h) uses `min_count = 1`, "emit null" below that many non-null values) and [PyArrow's `None`](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ScalarAggregateOptions.html)

Worth calling out, to remain consistent with Arrow, `Sum` also follow that same pattern now.

### Package Naming / Guidelines

Looking again at the other languages, I think the main reasoning for using those names and separate packages is consistency and following of established patterns in the broader Arrow ecosystem.

- **Core** - Describes the common data types and what the data is (buffers, schema, arrays, etc)
- **Compute** - Derives data from values (aggregate, sort, filter, arithmetic, etc.)

Naming aside, one other point to support the split, `Compute` adds the `System.Numerics.Tensors` dependency. While lightweight, not everyone may want to have that dependency in the box when they consume `Core` especially when they're not utilizing the *compute* operations.

cc: @CurtHagenlocher @adamreeve for visibility.

src/Apache.Arrow.Compute/Aggregations.cs
```diff
@@ -0,0 +1,407 @@
...
+{
+    /// <summary>
+    /// Aggregation kernels over <see cref="PrimitiveArray{T}"/> (Sum/Min/Max/Mean).
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Null entries are skipped and do not contribute to the result. <c>Sum</c>, <c>Min</c>,
+    /// <c>Max</c> and <c>Mean</c> return <c>null</c> (<see cref="System.Nullable{T}"/>) when the
+    /// array is empty or contains no non-null elements, matching Arrow's <c>ScalarAggregateOptions</c>
+    /// default (<c>min_count = 1</c>) and PyArrow's compute module, which return null in these cases.
```
adamreeve: It's not clear here that "Arrow" is actually referring to the C++ implementation. I think this is probably more detail than is needed though, this part of the sentence could just be removed:
```suggestion
    /// array is empty or contains no non-null elements.
```
Or if you'd like to keep this, it should say "C++ Arrow"

adamreeve (commented):
This looks good to me now thanks @luisquintanilla, just one small nitpick.

luisquintanilla:
Thanks @adamreeve. I applied your suggestion.

adamreeve (approved):

test/Apache.Arrow.Compute.Tests/Apache.Arrow.Compute.Tests.csproj
```diff
@@ -0,0 +1,19 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
CurtHagenlocher: Please run tests under net472 as well to get test coverage for the non-accelerated path.

src/Apache.Arrow.Compute/Aggregations.cs
```diff
@@ -0,0 +1,406 @@
...
+                return TensorPrimitives.Min(values);
+            }
+
+            T min = T.MaxValue;
+            for (int i = 0; i < values.Length; i++)
+            {
+                if (!array.IsValid(i)) continue;
+                if (values[i] < min) { min = values[i]; }
+            }
+            return min;
```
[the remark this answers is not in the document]
adamreeve: Ah my bad, I suggested this change, but Copilot is right.

CurtHagenlocher (commented):
Looks good! Can you just add net472 to the targets for the new test project?

EDIT: I forgot to check the Copilot feedback that I asked for. I think its points are good.

PR #393, opened 2026-07-20
feat: remove most calls to GC memory pressure APIs

### LINKED ISSUE #50: NativeMemoryAllocator doesn't remove all memory pressure

The `NativeMemoryAllocator` always over-allocates by the alignment amount to ensure the final buffer is aligned. When it [adds GC pressure](https://github.com/apache/arrow-dotnet/blob/710f6e47bd20c5a0dca471a27d67c2fff65d581c/src/Apache.Arrow/Memory/NativeMemoryAllocator.cs#L45) via `GC.AddMemoryPressure()`, this extra padding is accounted for. However, when `NativeAllocationOwner` [releases the memory allocation](https://github.com/apache/arrow-dotnet/blob/710f6e47bd20c5a0dca471a27d67c2fff65d581c/src/Apache.Arrow/Memory/NativeMemoryAllocator.cs#L58), it doesn't include the extra padding when it calls `GC.RemoveMemoryPressure()` (the `length` parameter is the size of the `Memory<byte>`, not the unmanaged allocation).

This doesn't appear to have long-term consequences, since the GC will periodically clear its unmanaged memory pressure counters, but my understanding is it can lead to extra garbage collections overall.

A tool like [`PerfView`](https://github.com/microsoft/perfview) can show events for each call to `AddMemoryPressure` and `RemoveMemoryPressure`. Running it with the following code shows a call to add memory pressure of 74 bytes and a call to remove memory pressure of 10 bytes:

```csharp
using Apache.Arrow.Memory;

MemoryAllocator.Default.Value.Allocate(10).Dispose();
```

### LINKED ISSUE #261: Consider batching calls to GC.AddMemoryPressure

In an internal project where I'm looking at the overhead of translating JSON responses into Arrow, I'm seeing the following performance characteristics:

.NET 4.7.2

| Method         | Mean     | Error   | StdDev  | Gen0      | Gen1     | Gen2     | Allocated |
|--------------- |---------:|--------:|--------:|----------:|---------:|---------:|----------:|
| SystemTextJson | 663.9 us | 3.32 us | 3.10 us |   58.5938 |  19.5313 |        - | 366.89 KB |
| Arrow          | 666.3 us | 5.77 us | 5.11 us | 7756.8359 | 965.8203 | 117.1875 |  97.06 KB |

.NET 8.0

| Method         | Mean     | Error   | StdDev  | Gen0     | Gen1     | Gen2     | Allocated |
|--------------- |---------:|--------:|--------:|---------:|---------:|---------:|----------:|
| SystemTextJson | 285.6 us | 1.00 us | 0.94 us |  23.4375 |  11.7188 |        - | 363.92 KB |
| Arrow          | 334.2 us | 6.46 us | 6.34 us | 199.7070 | 199.7070 | 199.7070 |  52.17 KB |

.NET 10.0

| Method         | Mean     | Error   | StdDev  | Gen0     | Gen1     | Gen2     | Allocated |
|--------------- |---------:|--------:|--------:|---------:|---------:|---------:|----------:|
| SystemTextJson | 267.2 us | 1.42 us | 1.26 us |  23.4375 |  14.1602 |        - | 362.37 KB |
| Arrow          | 331.9 us | 6.23 us | 5.83 us | 198.7305 | 198.7305 | 198.7305 |  48.44 KB |

When I effectively remove the calls to `GC.AddMemoryPressure` and `GC.RemoveMemoryPressure` but make no other changes, I instead get the following:

.NET 4.7.2

| Method         | Mean     | Error   | StdDev  | Gen0    | Gen1    | Allocated |
|--------------- |---------:|--------:|--------:|--------:|--------:|----------:|
| SystemTextJson | 661.6 us | 3.45 us | 3.23 us | 58.5938 | 19.5313 | 366.91 KB |
| Arrow          | 453.6 us | 8.35 us | 7.81 us |  7.8125 |  3.9063 |  53.68 KB |

.NET 8.0

| Method         | Mean     | Error   | StdDev  | Gen0    | Gen1    | Allocated |
|--------------- |---------:|--------:|--------:|--------:|--------:|----------:|
| SystemTextJson | 288.0 us | 2.57 us | 2.41 us | 23.4375 | 11.7188 | 363.92 KB |
| Arrow          | 277.9 us | 4.47 us | 3.73 us |  2.9297 |  2.4414 |  52.11 KB |

.NET 10.0

| Method         | Mean     | Error   | StdDev  | Gen0    | Gen1    | Allocated |
|--------------- |---------:|--------:|--------:|--------:|--------:|----------:|
| SystemTextJson | 270.4 us | 1.77 us | 1.66 us | 23.4375 | 14.1602 | 362.37 KB |
| Arrow          | 274.6 us | 3.46 us | 3.24 us |  2.9297 |  2.4414 |   48.3 KB |

The amount allocated by the "SystemTextJson" test is a reasonable proxy for the amount of native memory allocated by the "Arrow" test. Especially under .NET 4.7.2, it appears that these calls are triggering frequent garbage collection which (at least in a microbenchmark) are having a large effect on performance.

One way to deal with this is to effectively batch the calls. For instance, we might consider rounding up to 64 KB and only adding or removing memory pressure when we go past that threshold. For instance, the first allocation of 128 bytes would add 64 KB, then subsequent allocations of 128 bytes, 4096 bytes, 8192 bytes and 32768 bytes would only increment an internal counter. At that point, an allocation of 40000 bytes would push us past the next 64 KB boundary and trigger another `GC.AddMemoryPressure` for another 64 KB.

It's possible, of course, that the benchmark is pathological due to an overall heap size which is quite small at that point and that in a "real" application we would not see this kind of noise.  I'll try to figure out a way to measure that.

teo-tsirpanis:
## What's Changed

According to [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.gc.addmemorypressure?view=net-10.0#remarks) (see also https://github.com/dotnet/runtime/issues/33812#issuecomment-601733199):

> The [AddMemoryPressure](https://learn.microsoft.com/en-us/dotnet/api/system.gc.addmemorypressure?view=net-10.0) and [RemoveMemoryPressure](https://learn.microsoft.com/en-us/dotnet/api/system.gc.removememorypressure?view=net-10.0) methods improve performance only for types that exclusively depend on finalizers to release the unmanaged resources. It's not necessary to use these methods in types that follow the dispose pattern, where finalizers are used to clean up unmanaged resources only in the event that a consumer of the type forgets to call `Dispose`. […]

None of the uses of these methods by Arrow satisfies the above criterion. Consequently, this PR removes them, with the exception of `MemoryPressureAllocationTracker` whose public API contract is to call these methods, and which got obsoleted by this PR.

Closes #50
Closes #261

### DIFF
--- a/src/Apache.Arrow/Memory/ExportedAllocationOwner.cs
+++ b/src/Apache.Arrow/Memory/ExportedAllocationOwner.cs
@@ -26,7 +26,6 @@ namespace Apache.Arrow.Memory
         private readonly List<IntPtr> _pointers = new List<IntPtr>();
         private readonly List<MemoryHandle> _handles = new List<MemoryHandle>();
         private readonly List<SharedMemoryHandle> _sharedHandles = new List<SharedMemoryHandle>();
-        private long _allocationSize;
         private long _referenceCount;
         private bool _disposed;
 
@@ -37,14 +36,12 @@ namespace Apache.Arrow.Memory
 
         public IntPtr Allocate(int size)
         {
-            GC.AddMemoryPressure(size);
             return Acquire(Marshal.AllocHGlobal(size), 0, size);
         }
 
         public IntPtr Acquire(IntPtr ptr, int offset, int length)
         {
             _pointers.Add(ptr);
-            _allocationSize += length;
             return ptr;
         }
 
@@ -103,7 +100,6 @@ namespace Apache.Arrow.Memory
                 _sharedHandles[i] = default;
             }
 
-            GC.RemoveMemoryPressure(_allocationSize);
             GC.SuppressFinalize(this);
             _disposed = true;
         }
--- a/src/Apache.Arrow/Memory/ImportedAllocationOwner.cs
+++ b/src/Apache.Arrow/Memory/ImportedAllocationOwner.cs
@@ -41,7 +41,6 @@ namespace Apache.Arrow.Memory
             if (length > 0)
             {
                 Interlocked.Add(ref _managedMemory, length);
-                GC.AddMemoryPressure(length);
             }
             return memory;
         }
@@ -55,10 +54,6 @@ namespace Apache.Arrow.Memory
         {
             if (Interlocked.Decrement(ref _referenceCount) == 0)
             {
-                if (_managedMemory > 0)
-                {
-                    GC.RemoveMemoryPressure(_managedMemory);
-                }
                 FinalRelease();
             }
         }
--- a/src/Apache.Arrow/Memory/MemoryPressureAllocationTracker.cs
+++ b/src/Apache.Arrow/Memory/MemoryPressureAllocationTracker.cs
@@ -20,7 +20,8 @@ namespace Apache.Arrow.Memory
     /// <summary>
     /// Allows control over the way native allocations interact with the GC.
     /// </summary>
-    public struct MemoryPressureAllocationTracker : INativeAllocationTracker
+    [Obsolete("Use NoOpAllocationTracker instead. See https://learn.microsoft.com/en-us/dotnet/api/system.gc.addmemorypressure?view=net-10.0#remarks for more information on why the GC memory pressure APIs are not always necessary.")]
+    public readonly struct MemoryPressureAllocationTracker : INativeAllocationTracker
     {
         public void Track(int count, long bytes)
         {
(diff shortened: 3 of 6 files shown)


CurtHagenlocher:
I don't think this is a good idea without some other changes, particularly in the light of #397. As per @kronic's suggestion I'm looking at trying to do something more pluggable and have been prototyping some ideas.

PR #400, opened 2026-08-04
StringViewArray: enable nullable reference types and optimize encoding with stackalloc/ArrayPool

kronic:
StringViewArray: Enable nullable reference types and improve encoding performance

- Enable #nullable enable and update interfaces to use nullable string (IReadOnlyList<string?>)
- Change DefaultEncoding to UTF8Encoding(false) to explicitly disable BOM
- In Builder.Append, use GetMaxByteCount and stackalloc (for <=1024 bytes) or ArrayPool<byte>
  to avoid per‑string byte array allocation and reduce GC pressure
- Add using System.Buffers for ArrayPool
- Update Append and AppendRange to accept nullable strings and optional encoding
- Update GetString to return nullable string and accept optional encoding
- Use primary constructor for StringViewArray and Builder
- Adjust overloads and parameter defaults accordingly

### DIFF
--- a/src/Apache.Arrow/Arrays/StringViewArray.cs
+++ b/src/Apache.Arrow/Arrays/StringViewArray.cs
@@ -13,7 +13,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#nullable enable
+
 using System;
+using System.Buffers;
 using System.Collections;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
@@ -22,33 +25,56 @@ using Apache.Arrow.Types;
 
 namespace Apache.Arrow
 {
-    public class StringViewArray : BinaryViewArray, IReadOnlyList<string>
+    public class StringViewArray(ArrayData data) : BinaryViewArray(ArrowTypeId.StringView, data), IReadOnlyList<string?>
     {
-        public static readonly Encoding DefaultEncoding = Encoding.UTF8;
+        public static Encoding DefaultEncoding { get; } = new UTF8Encoding(false);
 
-        public new class Builder : BuilderBase<StringViewArray, Builder>
+        public new class Builder() : BuilderBase<StringViewArray, Builder>(StringViewType.Default)
         {
-            public Builder() : base(StringViewType.Default) { }
-
             protected override StringViewArray Build(ArrayData data)
             {
                 return new StringViewArray(data);
             }
 
-            public Builder Append(string value, Encoding encoding = null)
+            public Builder Append(string? value, Encoding? encoding = null)
             {
-                if (value == null)
+                if (value is null)
                 {
                     return AppendNull();
                 }
-                encoding = encoding ?? DefaultEncoding;
-                byte[] span = encoding.GetBytes(value);
-                return Append(span.AsSpan());
+
+                encoding ??= DefaultEncoding;
+                int maxByteCount = encoding.GetMaxByteCount(value.Length);
+                #if NETCOREAPP
+                byte[]? buffer = null;
+
+                Span<byte> span = maxByteCount <= 1024
+                    ? stackalloc byte[maxByteCount]
+                    : buffer = ArrayPool<byte>.Shared.Rent(maxByteCount);
+
+                int encodeBbytes = encoding.GetBytes(value, span);
+                span = span.Slice(0, encodeBbytes);
+                #else
+                byte[] buffer = ArrayPool<byte>.Shared.Rent(maxByteCount);
+                int encodeBbytes = encoding.GetBytes(value, 0, value.Length, buffer, 0);
+                Span<byte> span = buffer.AsSpan(0, encodeBbytes);
+                #endif
+                try
+                {
+                    return Append(span);
+                }
+                finally
+                {
+                    if (buffer != null)
+                    {
+                        ArrayPool<byte>.Shared.Return(buffer);
+                    }
+                }
             }
 
-            public Builder AppendRange(IEnumerable<string> values, Encoding encoding = null)
+            public Builder AppendRange(IEnumerable<string?> values, Encoding? encoding = null)
             {
-                foreach (string value in values)
+                foreach (string? value in values)
                 {
                     Append(value, encoding);
                 }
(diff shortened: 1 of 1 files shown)


src/Apache.Arrow/Arrays/StringViewArray.cs
```diff
@@ -22,33 +25,56 @@
 
 namespace Apache.Arrow
 {
-    public class StringViewArray : BinaryViewArray, IReadOnlyList<string>
+    public class StringViewArray(ArrayData data) : BinaryViewArray(ArrowTypeId.StringView, data), IReadOnlyList<string?>
     {
-        public static readonly Encoding DefaultEncoding = Encoding.UTF8;
+        public static Encoding DefaultEncoding { get; } = new UTF8Encoding(false);
```
adamreeve, line 30: I'm a little concerned that this could be a breaking behaviour change, but I checked that decoding is forgiving of whether or not a BOM is present, regardless of the `encoderShouldEmitUTF8Identifier` parameter. So I think this is OK.

I notice that we disable testing the `StringView` type in our C Data Interface Python integration tests: https://github.com/apache/arrow-dotnet/blob/dc2c56617f5612cfd77750ba79b21f0e9a2ec5e6/test/Apache.Arrow.Tests/CDataInterfacePythonTests.cs#L772-L773

Did you find that emitting the BOM was causing a problem when integrating with another Arrow implementation?

src/Apache.Arrow/Arrays/StringViewArray.cs
```diff
@@ -22,33 +25,56 @@
...
+                encoding ??= DefaultEncoding;
+                int maxByteCount = encoding.GetMaxByteCount(value.Length);
+                #if NETCOREAPP
+                byte[]? buffer = null;
+
+                Span<byte> span = maxByteCount <= 1024
+                    ? stackalloc byte[maxByteCount]
+                    : buffer = ArrayPool<byte>.Shared.Rent(maxByteCount);
+
+                int encodeBbytes = encoding.GetBytes(value, span);
```
adamreeve: `encoding.GetBytes` could throw, depending on the encoding used, so this should probably be within the `try` block.

src/Apache.Arrow/Arrays/StringViewArray.cs
kronic: @adamreeve Yes, BOM causes problems with DataFusion.

src/Apache.Arrow/Arrays/StringViewArray.cs
kronic: done

kronic:
In general, I think the idea of ​​temporary buffers is redundant. You can write directly to the ArrowBuffer, which will eliminate unnecessary memory copying and temporary buffer allocation.

src/Apache.Arrow/Arrays/StringViewArray.cs
```diff
@@ -22,33 +25,63 @@
...
-                encoding = encoding ?? DefaultEncoding;
-                byte[] span = encoding.GetBytes(value);
-                return Append(span.AsSpan());
+
+                encoding ??= DefaultEncoding;
+                int maxByteCount = encoding.GetMaxByteCount(value.Length);
+                #if NETCOREAPP
+                byte[]? buffer = null;
+
+                Span<byte> span = maxByteCount <= 1024
```
CurtHagenlocher, line 51: I've been using 256 as a conservative upper bound for stackalloc in libraries; see `VariantValueWriter.StackAllocThreshold`.

src/Apache.Arrow/Arrays/StringViewArray.cs
```diff
@@ -22,33 +25,63 @@
 
 namespace Apache.Arrow
 {
-    public class StringViewArray : BinaryViewArray, IReadOnlyList<string>
+    public class StringViewArray(ArrayData data) : BinaryViewArray(ArrowTypeId.StringView, data), IReadOnlyList<string?>
```
CurtHagenlocher, line 28: I think a codebase is easier to read when it's consistent in its use of syntax and personally found this new form jarring. I'm curious what other people think.

src/Apache.Arrow/Arrays/StringViewArray.cs
```diff
@@ -22,33 +25,63 @@
...
                 {
                     return AppendNull();
                 }
-                encoding = encoding ?? DefaultEncoding;
-                byte[] span = encoding.GetBytes(value);
-                return Append(span.AsSpan());
+
+                encoding ??= DefaultEncoding;
+                int maxByteCount = encoding.GetMaxByteCount(value.Length);
+                #if NETCOREAPP
```
CurtHagenlocher, line 48: Please put preprocessor directives in flush-left like the rest of the codebase.

CurtHagenlocher:
In general and for a variety of reasons I think the existing builders kind of suck. I've often been skipping them and building buffers directly for better performance. For your own use cases, do you find that you typically have a known set of .NET `string`s that you want to convert into an Arrow array or are you also building them more incrementally?

src/Apache.Arrow/Arrays/StringViewArray.cs
adamreeve: I also found it a bit jarring but I haven't been doing as much work in C# lately so maybe I'm just not used to this. I noticed Rider always suggests refactoring to this form.

src/Apache.Arrow/Arrays/StringViewArray.cs
adamreeve: In that case, I wonder if we should also change the [default for `StringArray`](https://github.com/apache/arrow-dotnet/blob/dc2c56617f5612cfd77750ba79b21f0e9a2ec5e6/src/Apache.Arrow/Arrays/StringArray.cs#L27). That could be slightly more risky given it's been around longer and is likely to be more widely used, but it would make sense to fix both at the same time.

Do you have an opinion on this @CurtHagenlocher?

CurtHagenlocher: To the best of my ability to tell, the BOM is only used in certain specific contexts such as `StreamWriter`. It is not referenced from `GetByteCount` or `GetBytes` and so should have no impact on the scenarios we care about.

CurtHagenlocher: That is, the following code prints `true`:
```
            string test = "test";
            byte[] one = new System.Text.UTF8Encoding(true).GetBytes(test);
            byte[] two = new System.Text.UTF8Encoding(false).GetBytes(test);
            System.Console.WriteLine(one.Length == two.Length);
```

adamreeve: OK, I'm confused about how this could have caused problems with DataFusion then. We only call `GetBytes` and `GetString` on this encoding. I guess that means this change is harmless, but I'd like to understand what motivated it then.

Issue #397, opened 2026-07-27
ArrowBuffer spans do not keep native memory alive: NativeMemoryManager's finalizer can free a buffer while a Span over it is in use


CurtHagenlocher:
### Summary

A `ReadOnlySpan<byte>` obtained from `ArrowBuffer.Span` does not keep the buffer's memory alive. Because `MemoryAllocator.Default` allocates native memory owned by a `NativeMemoryManager` **whose finalizer frees it**, an array that becomes unreachable while spans over its buffers are still in use can have those buffers freed underneath them.

The result is a use-after-free in ordinary-looking consumer code. The failure is usually silent — freed `HGlobal` pages typically stay mapped, so the span keeps returning plausible data — and only occasionally surfaces as an `AccessViolationException`.

### Reproduction

`repro.csproj`:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Apache.Arrow" Version="23.0.0" />
  </ItemGroup>
</Project>
```

`Program.cs`:

```csharp
using System.Buffers;
using System.Runtime.CompilerServices;
using System.Runtime.InteropServices;
using Apache.Arrow;

static StringArray BuildArray(int rows)
{
    var b = new StringArray.Builder();
    for (int i = 0; i < rows; i++) b.Append($"row-{i}");
    return b.Build();
}

// WeakReference to the MemoryManager that owns the array's value buffer.
static WeakReference OwnerOf(StringArray a)
{
    MemoryMarshal.TryGetMemoryManager<byte, MemoryManager<byte>>(
        a.Data.Buffers[2].Memory, out var manager);
    Console.WriteLine($"value buffer is owned by: {manager?.GetType().Name ?? "(managed array)"}");
    return new WeakReference(manager);
}

// Stands in for the allocating work a consumer does while holding the spans.
static void DoWorkWhileHoldingSpans()
{
    GC.Collect();
    GC.WaitForPendingFinalizers();
    GC.Collect();
}

[MethodImpl(MethodImplOptions.NoInlining)]
static bool WithoutKeepAlive()
{
    var array = BuildArray(100_000);
    var owner = OwnerOf(array);

[...]

PR #406, opened 2026-08-11
fix: Add memory lifetime warnings to ReadOnlySpan properties

Abhilash437:
## What's Changed

Adds explicit XML documentation remarks (`<remarks>`) to properties and methods that return a `ReadOnlySpan` over managed or unmanaged memory across Apache Arrow array and buffer classes.

### Motivation & Relation to PR #393
When an `ArrowBuffer` or `Array` is backed by unmanaged memory (e.g., allocated via a custom `MemoryAllocator` or `NativeMemoryManager`), extracting a `ReadOnlySpan` into a local variable and subsequently disposing the parent buffer/array can lead to use-after-free memory access issues.

While PR #393 explores structural lifecycle/pinning mechanisms, this PR provides an immediate, non-breaking developer safety improvement:
- Added `<remarks>` warning callouts to `ArrowBuffer.Span`, `PrimitiveArray.Values`, `BooleanArray.Values`, and `BinaryArray.GetBytes`. These surface as visible safety notes in IDE tooltips and IntelliSense (Visual Studio, VS Code, Rider) without breaking builds or requiring API breaking changes under `TreatWarningsAsErrors=true`.
- Added `PoisonMemoryAllocator` and unit test `TestNativeMemoryManagerUseAfterFree` to `Apache.Arrow.Tests` to verify memory poisoning on buffer release.

Relates to #397 and #393.

### DIFF
--- a/test/Apache.Arrow.Tests/ArrowBufferTests.cs
+++ b/test/Apache.Arrow.Tests/ArrowBufferTests.cs
@@ -116,7 +116,7 @@ namespace Apache.Arrow.Tests
         [Fact]
         public void TestNativeMemoryManagerUseAfterFree()
         {
-            var allocator = new PoisonMemoryAllocator();
+            using var allocator = new PoisonMemoryAllocator();
             // Allocate using the Builder pattern
             var builder = new ArrowBuffer.Builder<byte>(100000);
             builder.Append(new byte[100000]);
@@ -128,7 +128,7 @@ namespace Apache.Arrow.Tests
             // Dispose the buffer to trigger memory poisoning and release
             buffer.Dispose();
 
-            // This will now fail because span[50000] is poisoned with 0xFF after free instead of 0!
+            // span[50000] is poisoned with 0xFF after free, it's not the initial 0 value
             byte b = span[50000];
             Assert.Equal(0xFF, b);
         }
--- a/test/Apache.Arrow.Tests/ArrowBufferTests.cs
+++ b/test/Apache.Arrow.Tests/ArrowBufferTests.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the Apache Software Foundation (ASF) under one or more
+// Licensed to the Apache Software Foundation (ASF) under one or more
 // contributor license agreements. See the NOTICE file distributed with
 // this work for additional information regarding copyright ownership.
 // The ASF licenses this file to You under the Apache License, Version 2.0
@@ -14,7 +14,9 @@
 // limitations under the License.
 
 using System;
+using System.Buffers;
 using System.Threading;
+using Apache.Arrow.Memory;
 using Apache.Arrow.Tests.Fixtures;
 using Xunit;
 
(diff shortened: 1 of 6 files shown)


test/Apache.Arrow.Tests/PoisonMemoryAllocator.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            public void Dispose()
+            {
+                if (_disposed)
+                {
+                    return;
+                }
+                _disposed = true;
+                _inner.Memory.Span.Fill(0xFF);
+                _inner.Dispose();
```
adamreeve: Won't this end up calling `Marshal.FreeHGlobal` on the allocated memory?
https://github.com/apache/arrow-dotnet/blob/c42689e99046fe5709492f2fcdd9ceafa937bfad/src/Apache.Arrow/Memory/NativeMemoryAllocator.cs#L57

As per the comment in #397:
> The failure is usually silent — freed HGlobal pages typically stay mapped, so the span keeps returning plausible data — and only occasionally surfaces as an AccessViolationException.

So reading from the Span above reads from freed memory and is not safe. You might want to instead keep the inner `IMemoryOwner` undisposed and alive so you can test the poisoning safely?

test/Apache.Arrow.Tests/ArrowBufferTests.cs
```diff
@@ -111,6 +113,26 @@ public void TestExternalMemoryWrappedAsArrowBuffer()
...
+            builder.Append(new byte[100000]);
+            ArrowBuffer buffer = builder.Build(allocator);
+
+            // Extract the unmanaged Span
+            ReadOnlySpan<byte> span = buffer.Span;
+
+            // Dispose the buffer to trigger memory poisoning and release
+            buffer.Dispose();
+
+            // This will now fail because span[50000] is poisoned with 0xFF after free instead of 0!
```
adamreeve: This comment seems wrong to me. Won't the assertion pass because the value is 0xFF due to the poisoning?
```suggestion
            // span[50000] is poisoned with 0xFF after free, it's not the initial 0 value
```

adamreeve:
I've removed "closes #397" from your PR description. I think we want to keep that issue open to explore other solutions.

test/Apache.Arrow.Tests/ArrowBufferTests.cs
Abhilash437: updated the comment

test/Apache.Arrow.Tests/PoisonMemoryAllocator.cs
Abhilash437: Updated `PoisonMemoryAllocator` to implement `IDisposable` and track all inner `IMemoryOwner` instances. `PoisonMemoryOwner.Dispose()` now poisons the buffer (`0xFF`) without calling `_inner.Dispose()`, and the underlying native memory is cleanly freed when `PoisonMemoryAllocator.Dispose()` runs at the end of the test scope via `using var allocator`.

adamreeve (approved):

PR #408, opened 2026-08-11
feat: Implement RunEndEncodedArray.Builder (#407)

### LINKED ISSUE #407: Implement Array Builder for RunEndEncodedArray
Currently, the `RunEndEncodedArray` type in `Apache.Arrow.Arrays` lacks an implementation for `IArrowArrayBuilder`. 

Run-End Encoded arrays are widely used for compressing sequences of repeated values, but users currently have no easy way to build them programmatically in C#. We should add a `RunEndEncodedArray.Builder` class that handles the implicit tracking of run-lengths (using an `EqualityComparer`) and routes the compressed data to the underlying child arrays for the run-ends and values.

This will bring the C# implementation up to parity with the Arrow specification for constructing REE arrays.

Abhilash437:
## What's Changed

This PR implements `RunEndEncodedArray.Builder` to close the feature gap for constructing REE arrays in C#. 

- Designed the Builder to accept generic `TRunEndBuilder` and `TValueBuilder` types in the constructor (pre-configured inner builders).
- Implemented implicit run-length tracking via `Append(TValue value)`, utilizing `EqualityComparer<TValue>.Default` to group sequential duplicates.
- Added comprehensive unit tests in `RunEndEncodedArrayBuilderTests.cs`.

Closes #407.

### DIFF
--- a/src/Apache.Arrow/Arrays/RunEndEncodedArray.Builder.cs
+++ b/src/Apache.Arrow/Arrays/RunEndEncodedArray.Builder.cs
@@ -60,7 +60,7 @@ public partial class RunEndEncodedArray
         /// <summary>
         /// Initializes a new instance of the <see cref="Builder{TRunEndBuilder, TValueBuilder, TRunEndArray, TValueArray, TValue}"/> class.
         /// </summary>
-        /// <param name="runEndsBuilder">The builder to use for run-ends.</param>
+        /// <param name="runEndsBuilder">The builder to use for run-ends. Must be a builder for an Int16, Int32 or Int64 array.</param>
         /// <param name="valuesBuilder">The builder to use for values.</param>
         /// <param name="comparer">Optional equality comparer for value run-length grouping.</param>
         public Builder(TRunEndBuilder runEndsBuilder, TValueBuilder valuesBuilder, IEqualityComparer<TValue> comparer = null)
@@ -86,7 +86,10 @@ public partial class RunEndEncodedArray
             {
                 if (!_lastValueIsNull && _comparer.Equals(value, _lastValue))
                 {
-                    _length++;
+                    checked
+                    {
+                        _length++;
+                    }
                 }
                 else
                 {
@@ -112,7 +115,10 @@ public partial class RunEndEncodedArray
             {
                 if (_lastValueIsNull)
                 {
-                    _length++;
+                    checked
+                    {
+                        _length++;
+                    }
                 }
                 else
                 {
@@ -162,9 +168,9 @@ public partial class RunEndEncodedArray
         }
 
         /// <summary>
-        /// Reserves capacity in the inner builders.
+        /// Validates the capacity argument. Does not preallocate inner builders to allow doubling growth strategy with REE compression.
         /// </summary>
-        /// <param name="capacity">The capacity to reserve.</param>
+        /// <param name="capacity">The capacity to validate.</param>
         /// <returns>The builder instance for method chaining.</returns>
         public Builder<TRunEndBuilder, TValueBuilder, TRunEndArray, TValueArray, TValue> Reserve(int capacity)
         {
@@ -173,8 +179,6 @@ public partial class RunEndEncodedArray
                 throw new ArgumentOutOfRangeException(nameof(capacity));
             }
 
-            ReserveBuilder<TRunEndArray, TRunEndBuilder>(RunEndsBuilder, capacity);
-            ReserveBuilder<TValueArray, TValueBuilder>(ValuesBuilder, capacity);
             return this;
         }
 
@@ -242,7 +246,10 @@ public partial class RunEndEncodedArray
             _lastValue = value;
             _lastValueIsNull = isNull;
             _hasValue = true;
-            _length++;
+            checked
+            {
+                _length++;
+            }
         }
 
         private void FlushCurrentRun()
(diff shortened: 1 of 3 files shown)


src/Apache.Arrow/Arrays/RunEndEncodedArray.Builder.cs
```diff
@@ -0,0 +1,400 @@
...
+        where TRunEndBuilder : IArrowArrayBuilder<TRunEndArray>
+        where TValueBuilder : IArrowArrayBuilder<TValueArray>
+        where TRunEndArray : IArrowArray
+        where TValueArray : IArrowArray
+    {
+        private readonly IEqualityComparer<TValue> _comparer;
+        private int _currentRunLength;
+        private TValue _lastValue;
+        private bool _lastValueIsNull;
+        private bool _hasValue;
```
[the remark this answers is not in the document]
Abhilash437: Renamed private field `_currentRunLength` to `_length` across `RunEndEncodedArray.Builder.cs` to accurately reflect that it stores the cumulative logical length of the array.

src/Apache.Arrow/Arrays/RunEndEncodedArray.Builder.cs
```diff
@@ -0,0 +1,400 @@
...
+
+        /// <summary>
+        /// Reserves capacity in the inner builders.
+        /// </summary>
+        /// <param name="capacity">The capacity to reserve.</param>
+        /// <returns>The builder instance for method chaining.</returns>
+        public Builder<TRunEndBuilder, TValueBuilder, TRunEndArray, TValueArray, TValue> Reserve(int capacity)
+        {
+            return this;
+        }
```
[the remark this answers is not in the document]
Abhilash437: `Reserve(capacity)` now validates input and throws an `ArgumentOutOfRangeException` when `capacity < 0`. Capacity reservation is delegated to both inner builders (`RunEndsBuilder` and `ValuesBuilder`) via a new `ReserveBuilder` helper method.

src/Apache.Arrow/Arrays/RunEndEncodedArray.Builder.cs
```diff
@@ -0,0 +1,400 @@
...
+
+        /// <summary>
+        /// Resizes the builder.
+        /// </summary>
+        /// <param name="length">The target length.</param>
+        /// <returns>The builder instance for method chaining.</returns>
+        public Builder<TRunEndBuilder, TValueBuilder, TRunEndArray, TValueArray, TValue> Resize(int length)
+        {
+            return this;
+        }
```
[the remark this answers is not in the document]
Abhilash437: Updated `Resize(length)` to throw a `NotSupportedException("Resize is not supported on RunEndEncodedArray.Builder.")` to align with the behavior of `Set` and `Swap` on REE builders.

src/Apache.Arrow/Arrays/RunEndEncodedArray.Builder.cs
```diff
@@ -0,0 +1,400 @@
...
+            }
+            else if (builder is LargeBinaryArray.Builder lbb)
+            {
+                lbb.Clear();
+            }
+            else if (builder is BinaryViewArray.Builder bvb)
+            {
+                bvb.Clear();
+            }
+        }
```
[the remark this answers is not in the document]
Abhilash437: The fallback `else` branch of `ClearBuilder` now explicitly throws a `NotSupportedException` for unsupported builder types to fail fast and prevent silent state corruption.

src/Apache.Arrow/Arrays/RunEndEncodedArray.Builder.cs
```diff
@@ -0,0 +1,450 @@
...
+            if (value is null)
+            {
+                return AppendNull();
+            }
+
+            if (_hasValue)
+            {
+                if (!_lastValueIsNull && _comparer.Equals(value, _lastValue))
+                {
+                    _length++;
```
adamreeve: We should check for overflow on length increments:

```suggestion
                    checked {
                        _length++;
                    }
```

(and similarly for other places where we increment the length)

src/Apache.Arrow/Arrays/RunEndEncodedArray.Builder.cs
```diff
@@ -0,0 +1,450 @@
...
+        /// <returns>The builder instance for method chaining.</returns>
+        public Builder<TRunEndBuilder, TValueBuilder, TRunEndArray, TValueArray, TValue> Reserve(int capacity)
+        {
+            if (capacity < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(capacity));
+            }
+
+            ReserveBuilder<TRunEndArray, TRunEndBuilder>(RunEndsBuilder, capacity);
+            ReserveBuilder<TValueArray, TValueBuilder>(ValuesBuilder, capacity);
```
adamreeve: For data that compresses well with REE, this could end up reserving a lot more capacity than is actually needed. Maybe it should be a no-op instead, so we use the doubling growth strategy of the underlying buffer builders?

src/Apache.Arrow/Arrays/RunEndEncodedArray.Builder.cs
```diff
@@ -0,0 +1,450 @@
...
+
+        /// <summary>
+        /// Gets the total logical length of elements appended to this builder.
+        /// </summary>
+        public int Length => _length;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Builder{TRunEndBuilder, TValueBuilder, TRunEndArray, TValueArray, TValue}"/> class.
+        /// </summary>
+        /// <param name="runEndsBuilder">The builder to use for run-ends.</param>
```
adamreeve:
```suggestion
        /// <param name="runEndsBuilder">The builder to use for run-ends. Must be a builder for an Int16, Int32 or Int64 array.</param>
```

adamreeve (commented):
Thanks for contributing this @Abhilash437, I've left a few comments

src/Apache.Arrow/Arrays/RunEndEncodedArray.Builder.cs
Abhilash437: Applied the suggestion

src/Apache.Arrow/Arrays/RunEndEncodedArray.Builder.cs
Abhilash437: Agreed, I have kept the capacity check to throw the exception and made the function a no-op.

src/Apache.Arrow/Arrays/RunEndEncodedArray.Builder.cs
Abhilash437: Agreed, I have made similar changes for other places

adamreeve (approved):

PR #426, opened 2026-08-25
fix: Return immediately from a zero-length ReadFullBufferAsync/ReadFullBuffer instead of calling into the stream

### LINKED ISSUE #425: ArrowStreamReader hangs indefinitely reading a RecordBatch with a zero-length body over a NetworkStream
# `ArrowStreamReader` hangs indefinitely reading a `RecordBatch` with a zero-length body over a `NetworkStream`

## Summary

`ArrowStreamReaderImplementation.ReadMessageAsync` (and the sync `ReadMessage`) reads a record
batch's body via `StreamExtensions.ReadFullBufferAsync`/`ReadFullBuffer`, passing a
zero-length buffer whenever the batch has no buffers (e.g. a batch built from a zero-column
schema). Over a `MemoryStream` this is a harmless no-op. Over a real socket-backed
`NetworkStream`, a zero-byte `Stream.ReadAsync`/`Stream.Read` call does **not** complete
immediately — it blocks as though waiting for the peer to send more data (or close the
connection), instead of trivially returning `0`. In a lockstep/RPC-style protocol where the
peer is itself waiting for a response before sending anything further, this blocks forever.

## Repro

Minimal reproducer (no dependency on any RPC framework, just this package + real `pyarrow`
over a Unix domain socket):

**Server (C#, stock `Apache.Arrow` NuGet package, no vendored/forked code):**

```csharp
using System.Net.Sockets;
using Apache.Arrow;
using Apache.Arrow.Ipc;
using Apache.Arrow.Types;

var sockPath = args[0];
if (File.Exists(sockPath)) File.Delete(sockPath);

using var listener = new Socket(AddressFamily.Unix, SocketType.Stream, ProtocolType.Unspecified);
listener.Bind(new UnixDomainSocketEndPoint(sockPath));
listener.Listen();
Console.WriteLine("READY");

using var accepted = await listener.AcceptAsync();
var readStream = new NetworkStream(accepted, FileAccess.Read, ownsSocket: false);
var writeStream = new NetworkStream(accepted, FileAccess.Write, ownsSocket: false);

var outputSchema = new Schema([new Field("index", Int64Type.Default, nullable: false)], null);
var tickSchema = new Schema([], null);  // zero columns

using var writer = new ArrowStreamWriter(writeStream, outputSchema, leaveOpen: true);
await writer.WriteStartAsync();

using var reader = new ArrowStreamReader(readStream, leaveOpen: true);
_ = await reader.GetSchema();

var sw = System.Diagnostics.Stopwatch.StartNew();
var tick = await reader.ReadNextRecordBatchAsync();   // <-- blocks for ~8s (client's own timeout)
Console.WriteLine($"tick read in {sw.ElapsedMilliseconds}ms (rows={tick?.Length ?? -1})");
```

**Client (Python, real `pyarrow`, matching how a typical socket-backed IPC client is wired):**

```python
import pyarrow as pa
import pyarrow.ipc as ipc
import socket, sys

sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
sock.connect(sys.argv[1])
reader_stream = sock.makefile("rb")
writer_stream = sock.makefile("wb", buffering=0)   # unbuffered — makes no difference either way

tick_schema = pa.schema([])  # zero columns
tick_writer = ipc.new_stream(writer_stream, tick_schema)
output_reader = ipc.open_stream(reader_stream)

tick_batch = pa.record_batch([], schema=tick_schema)
tick_writer.write_batch(tick_batch)   # writes real bytes synchronously — confirmed via strace
                                       # and via a separate BytesIO capture; not the problem
response = output_reader.read_next_batch()
```

Result: the C# server's `ReadNextRecordBatchAsync()` call blocks for the full client-side
socket timeout (tested at 5s, 10s, and 30s — it always takes *exactly* as long as the client
is willing to wait, never less), even though the client wrote its complete message bytes
synchronously and promptly. Once the client gives up and closes its socket, the server's read
"completes" (returning `0` bytes, indistinguishable from an EOF), and the batch is then
constructed correctly and instantly — the delay is 100% attributable to the read call itself,
not any parsing or buffering elsewhere.

## Root cause (pinned via `strace`)

Running the server above under `strace -f -tt`, the sequence for the hung read is:

```
recvmsg(fd, ..., iov_len=4)   = 4     # continuation marker
recvmsg(fd, ..., iov_len=4)   = 4     # message length = 64
recvmsg(fd, ..., iov_len=64)  = 64    # full message body — arrives correctly and promptly
recvmsg(fd, MSG_PEEK)         = -1 EAGAIN
# ... nothing else happens on this fd for ~8 seconds ...
```

Adding timestamped instrumentation directly into `ArrowStreamReaderImplementation.ReadMessageAsync`
and `StreamExtensions.ReadFullBufferAsync` confirms exactly where the time goes: the 64-byte
*message* (the flatbuffer-encoded `Message` describing the batch) is read correctly and
immediately. `message.BodyLength` for this batch is `0` (zero columns → zero buffers → zero body
bytes needed). The subsequent call:

```csharp
IMemoryOwner<byte> bodyBuffOwner = AllocateMessageBodyBuffer(bodyLength);   // bodyLength == 0
Memory<byte> bodyBuff = bodyBuffOwner.Memory.Slice(0, bodyLength);          // 0-length slice
bytesRead = await BaseStream.ReadFullBufferAsync(bodyBuff, cancellationToken)...
```

drills down into:

```csharp
// StreamExtensions.ReadFullBufferAsync
int bytesRead = await stream.ReadAsync(buffer.Slice(...), cancellationToken)...
```

with `buffer.Length == 0`. This is the call that blocks for the full ~8 seconds. The
instrumented log shows it precisely:

```
DIAG-RFB [...746]: about to ReadAsync, want 0 more bytes (have 0/0)
DIAG-RFB [...689]: ReadAsync returned 0 bytes        <- ~7.9 seconds later
```

Feeding the exact same captured wire bytes through a `MemoryStream` instead of a real
`NetworkStream` parses correctly and instantly — `MemoryStream.ReadAsync` with a zero-length
buffer returns `0` immediately, as one would expect. `NetworkStream` (backed by a real
`Socket`) does not.

## Why this matters in practice

Any RPC/streaming protocol built on this library where a legitimate message can have a
zero-length body (e.g. a "tick"/"continue" message carrying no payload, a schema with no
columns, an empty control message) will hang the first time it's sent over a real socket
transport, even though the writer sent everything correctly and promptly. This is easy to miss
in testing because:
- `MemoryStream`-based tests never exercise the bug (zero-length reads are trivially
  synchronous there).
- Pipe-based transports may not exhibit it either (a hypothesis, not fully confirmed) since
  `FileStream`/anonymous-pipe reads for a zero-length buffer may behave differently than a
  `Socket`-backed `NetworkStream`.
- It only manifests over a real, separate-process, socket-backed connection.

## Suggested fix

`StreamExtensions.ReadFullBufferAsync` / `ReadFullBuffer` should short-circuit
`buffer.Length == 0` and return `0` immediately, without ever calling `stream.ReadAsync`/`Read`:

```csharp
public static async ValueTask<int> ReadFullBufferAsync(this Stream stream, Memory<byte> buffer, CancellationToken cancellationToken = default)
{
    if (buffer.Length == 0)
    {
        return 0;
    }

    int totalBytesRead = 0;
    // ... unchanged ...
}
```

This matches the Go standard library's documented behavior for `io.ReadFull`, which explicitly
returns immediately without issuing a read when `len(buf) == 0`, and sidesteps whatever
`NetworkStream`/`Socket.ReceiveAsync`-specific semantics cause a zero-length read to block on
readability rather than completing trivially.

We've applied this exact patch locally (in a vendored fork, since we also carry an unrelated
custom_metadata patch) and confirmed it fixes the hang completely: 5 consecutive request/response
turns using zero-column batches now complete in single-digit milliseconds over the same real
Unix-domain-socket setup that previously hung for the full duration of the peer's timeout, every
time.

## Environment

- `Apache.Arrow` NuGet package version: 18.1.0 (also reproduces on the version vendored from the
  `v23.0.0` tag)
- .NET 9 and .NET 10, reproduced on Linux x86_64, Linux ARM64, macOS, and Windows 11
- Reproduces regardless of whether the writer-side socket is buffered or unbuffered
  (`buffering=0`), and regardless of whether the reader uses split read/write `NetworkStream`
  instances or a single shared one

rustyconover:
## What's Changed

`StreamExtensions.ReadFullBufferAsync`/`ReadFullBuffer` called `stream.ReadAsync`/`stream.Read` with a zero-length buffer unconditionally, whenever a message body is legitimately empty (e.g. a `RecordBatch` built from a zero-column schema, which has no buffers). Over a `MemoryStream` this is a harmless no-op, but over a real socket-backed `NetworkStream` a zero-byte read does not complete immediately — it blocks as though waiting for the peer to send more data, instead of trivially returning `0`. In a lockstep/RPC-style protocol this blocks indefinitely, since the peer is itself waiting for a response before sending anything further.

Both helpers now short-circuit `buffer.Length == 0` and return `0` immediately, before ever touching the stream — matching Go's `io.ReadFull`, which documents the same special-case for a zero-length buffer.

Full repro (real `pyarrow` client, real socket, `strace` evidence pinning the exact hang to this call) is in the linked issue.

Added `StreamExtensionsTests.cs`: a `Stream` subclass whose `Read`/`ReadAsync` throw if ever invoked confirms the zero-length fast path never touches the underlying stream, plus two tests confirming normal non-empty reads are unaffected. `dotnet test test/Apache.Arrow.Tests` passes in full (1874 passed, 28 skipped — unrelated Python interop tests, pre-existing).

Closes #425.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/Apache.Arrow.Tests/StreamExtensionsTests.cs
+++ b/test/Apache.Arrow.Tests/StreamExtensionsTests.cs
@@ -48,8 +48,12 @@ namespace Apache.Arrow.Tests
             public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken) =>
                 throw new InvalidOperationException("ReadAsync should not be called for a zero-length buffer.");
 
+#if NETCOREAPP
+            // Stream.ReadAsync(Memory<byte>, CancellationToken) is only overridable on
+            // netcoreapp targets — net462/net472 don't declare it as virtual on Stream.
             public override ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationToken cancellationToken = default) =>
                 throw new InvalidOperationException("ReadAsync should not be called for a zero-length buffer.");
+#endif
 
             public override void Flush() => throw new NotSupportedException();
             public override long Seek(long offset, SeekOrigin origin) => throw new NotSupportedException();
--- a/test/Apache.Arrow.Tests/StreamExtensionsTests.cs
+++ b/test/Apache.Arrow.Tests/StreamExtensionsTests.cs
@@ -1,0 +1,98 @@
+// Licensed to the Apache Software Foundation (ASF) under one or more
+// contributor license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright ownership.
+// The ASF licenses this file to You under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance with
+// the License.  You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Apache.Arrow.Tests
+{
+    public class StreamExtensionsTests
+    {
+        /// <summary>
+        /// A stream whose Read/ReadAsync overrides throw if ever invoked, standing in for a
+        /// socket-backed stream (e.g. NetworkStream) whose zero-length ReadAsync/Read does not
+        /// complete immediately the way MemoryStream's does — it blocks as though waiting for the
+        /// peer to send more data. Used to prove ReadFullBufferAsync/ReadFullBuffer never call
+        /// into the underlying stream for a zero-length request.
+        /// </summary>
+        private sealed class ThrowsIfReadStream : Stream
+        {
+            public override bool CanRead => true;
+            public override bool CanSeek => false;
+            public override bool CanWrite => false;
+            public override long Length => throw new NotSupportedException();
+            public override long Position
+            {
+                get => throw new NotSupportedException();
+                set => throw new NotSupportedException();
+            }
+
+            public override int Read(byte[] buffer, int offset, int count) =>
+                throw new InvalidOperationException("Read should not be called for a zero-length buffer.");
+
+            public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken) =>
+                throw new InvalidOperationException("ReadAsync should not be called for a zero-length buffer.");
+
+            public override ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationToken cancellationToken = default) =>
+                throw new InvalidOperationException("ReadAsync should not be called for a zero-length buffer.");
+
+            public override void Flush() => throw new NotSupportedException();
+            public override long Seek(long offset, SeekOrigin origin) => throw new NotSupportedException();
+            public override void SetLength(long value) => throw new NotSupportedException();
+            public override void Write(byte[] buffer, int offset, int count) => throw new NotSupportedException();
+        }
+
+        [Fact]
+        public async Task ReadFullBufferAsync_ZeroLengthBuffer_ReturnsWithoutTouchingStream()
+        {
+            var stream = new ThrowsIfReadStream();
+            int bytesRead = await stream.ReadFullBufferAsync(Memory<byte>.Empty);
+            Assert.Equal(0, bytesRead);
+        }
+
+        [Fact]
+        public void ReadFullBuffer_ZeroLengthBuffer_ReturnsWithoutTouchingStream()
+        {
+            var stream = new ThrowsIfReadStream();
+            int bytesRead = stream.ReadFullBuffer(Memory<byte>.Empty);
+            Assert.Equal(0, bytesRead);
+        }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


test/Apache.Arrow.Tests/StreamExtensionsTests.cs
```diff
@@ -0,0 +1,98 @@
...
+                set => throw new NotSupportedException();
+            }
+
+            public override int Read(byte[] buffer, int offset, int count) =>
+                throw new InvalidOperationException("Read should not be called for a zero-length buffer.");
+
+            public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken) =>
+                throw new InvalidOperationException("ReadAsync should not be called for a zero-length buffer.");
+
+            public override ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationToken cancellationToken = default) =>
```
CurtHagenlocher, line 54: This will need an `#if` to exclude it from netstandard20 builds

CurtHagenlocher (commented):
Thanks! Other than the netstandard2.0 build break, looks good.

CurtHagenlocher (approved):

PR #283, opened 2026-03-11
Support IPC Message custom_metadata in ArrowStreamWriter and ArrowStreamReader  Body

cmettler:
Closes https://github.com/apache/arrow-dotnet/issues/282#issue-4058617524

PR #432, opened 2026-09-10
feat: Support IPC Message custom_metadata in ArrowStreamWriter and ArrowStreamReader

### LINKED ISSUE #282: Support IPC Message custom_metadata in ArrowStreamWriter and ArrowStreamReader  Body

The Arrow IPC [Message flatbuffer](https://github.com/apache/arrow/blob/main/format/Message.fbs#L36) defines a custom_metadata field ([KeyValue]) that allows attaching arbitrary key-value pairs to individual messages (record batches). The C# implementation currently ignores this field on both read and write.

Motivation
PyArrow has supported writing custom_metadata via write_batch(batch, custom_metadata=...) and reading it via read_next_batch_with_custom_metadata() since v11.0.0. The Rust implementation also added support recently (apache/arrow-rs#9444). The C# library should reach parity with other implementations.

This enables use cases such as:

RPC frameworks embedding method routing info in per-batch metadata
Logging/tracing metadata attached to individual batches
Application-level framing information in streaming IPC
Proposed Changes
Writer: Add WriteRecordBatch(RecordBatch, IReadOnlyDictionary<string, string>) overloads (sync + async) to ArrowStreamWriter that serialize the key-value pairs into the Message's custom_metadata field.

Reader: Expose a LastBatchCustomMetadata property on ArrowStreamReader that returns the custom_metadata from the most recently read batch's Message.

CurtHagenlocher:
Adds IPC `Message.custom_metadata` support to the stream and file readers and writers.

## Credit

The original implementation is **@cmettler's** work in #283 (closing #282). **@rustyconover** rebased it onto a `main` that had drifted ~76 commits ahead and addressed the first round of review comments in #424. Both of those PRs are branches on personal forks; this one moves the work onto a branch in the base repository so that any committer can push to it, and supersedes them. The individual commits here retain their original authorship.

## What this adds

- `ArrowStreamReader.ReadNextRecordBatchWithCustomMetadata()` and `…Async()` return a `RecordBatchWithMetadata` pairing the batch with its IPC `Message.custom_metadata`, mirroring pyarrow's `read_next_batch_with_custom_metadata()` and the equivalent Arrow C++ struct. The struct deconstructs, so callers can write `var (batch, metadata) = …`.
- `ArrowFileReader.ReadRecordBatchWithCustomMetadataAsync(int index)` gives the indexed read the same capability as the sequential one.
- `ArrowStreamWriter.WriteRecordBatch(batch, customMetadata)` and the async counterpart attach per-message `custom_metadata` when writing, matching pyarrow's `write_batch(batch, custom_metadata)`.
- Cross-language round-trip tests via pythonnet + pyarrow, skipped unless `PYTHONNET_PYDLL` is set, consistent with the existing `CDataSchemaPythonTest` pattern.

## Changes on top of #424

- The read side originally exposed a `LastBatchCustomMetadata` property. A property that has to be read at exactly the right moment is easy to get out of step with the batch in hand, and it had no sensible value at the end of the stream, so it was replaced with the `RecordBatchWithMetadata` return type above. The transient state on `ArrowReaderImplementation` remains, but it is internal and consumed immediately.
- The new write overloads went straight to `WriteRecordBatchInternal`, while `ArrowFileWriter` relied on overriding each public `WriteRecordBatch` to call `WriteStart()` first — so the new overload on an `ArrowFileWriter` skipped the ARROW1 magic and silently produced a file that `ArrowFileReader` rejects. `WriteStart()`/`WriteStartAsync()` moved into `WriteRecordBatchInternal`, where every write path must pass through it; both are idempotent, so byte output is unchanged for the stream writer, the file writer and Flight.
- Removed the second virtual `WriteMessageAsync` overload. Two virtual overloads where one forwards to the other is the trap that already routed Flight's record batch writes past `FlightDataStream`'s override.
- Custom metadata is validated before anything is written rather than part-way through building the message, so a rejected dictionary leaves the writer usable.
- `Message.custom_metadata` is read the same way schema and field metadata already are in `MessageSerializer`, rather than skipping null keys and rewriting null values as `""`.

## Verification

- `dotnet build Apache.Arrow.sln` — succeeds, 0 warnings, 0 errors.
- `dotnet test test/Apache.Arrow.Tests` — 1893 passed / 30 skipped on net8.0, 1849 passed / 30 skipped on net462 and net472, 0 failed. The skips are the pre-existing Python-dependent tests.

Closes #282
Supersedes #283
Supersedes #424

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DT86mdkGm3XseKwiUeGUcx

Co-Authored-By: Christoph Mettler <116812500+cmettler@users.noreply.github.com>
Co-Authored-By: Rusty Conover <rusty@conover.me>

### DIFF
--- a/src/Apache.Arrow.Flight/Internal/FlightDataStream.cs
+++ b/src/Apache.Arrow.Flight/Internal/FlightDataStream.cs
@@ -55,7 +55,7 @@ namespace Apache.Arrow.Flight.Internal
 
             var offset = SerializeSchema(Schema);
             CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
-            await WriteMessageAsync(MessageHeader.Schema, offset, 0, cancellationTokenSource.Token).ConfigureAwait(false);
+            await WriteMessageAsync(MessageHeader.Schema, offset, 0, default, cancellationTokenSource.Token).ConfigureAwait(false);
             await _clientStreamWriter.WriteAsync(_currentFlightData).ConfigureAwait(false);
             HasWrittenSchema = true;
         }
@@ -81,7 +81,7 @@ namespace Apache.Arrow.Flight.Internal
                 _currentFlightData.AppMetadata = applicationMetadata;
             }
 
-            await WriteRecordBatchInternalAsync(recordBatch).ConfigureAwait(false);
+            await WriteRecordBatchInternalAsync(recordBatch, customMetadata: null).ConfigureAwait(false);
 
             //Reset stream position
             this.BaseStream.Position = 0;
--- a/src/Apache.Arrow.Flight/Internal/FlightDataStream.cs
+++ b/src/Apache.Arrow.Flight/Internal/FlightDataStream.cs
@@ -91,11 +91,11 @@ namespace Apache.Arrow.Flight.Internal
             await _clientStreamWriter.WriteAsync(_currentFlightData).ConfigureAwait(false);
         }
 
-        private protected override ValueTask<long> WriteMessageAsync<T>(MessageHeader headerType, Offset<T> headerOffset, int bodyLength, CancellationToken cancellationToken)
+        private protected override ValueTask<long> WriteMessageAsync<T>(MessageHeader headerType, Offset<T> headerOffset, int bodyLength, VectorOffset customMetadataOffset, CancellationToken cancellationToken)
         {
             Offset<Flatbuf.Message> messageOffset = Flatbuf.Message.CreateMessage(
                 Builder, CurrentMetadataVersion, headerType, headerOffset.Value,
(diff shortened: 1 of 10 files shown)


rustyconover:
Thanks @CurtHagenlocher I've just been away from this for a little while.

CurtHagenlocher:
> Thanks @CurtHagenlocher I've just been away from this for a little while.

Sorry, I wouldn't have done this normally but there are too many partially-done things swimming around in my head right now :/.

rustyconover:
I get it, too many open mental windows.  It feels nice to close the mental tabs too.

Issue #180, opened 2025-11-18
Add support for sending dictionary arrays via Flight


CurtHagenlocher:
### Describe the enhancement requested

Dictionary arrays are not supported currently in dotnet flight implementation

Initially implement dictionary support where dictionary messages are resent with every record batch. This is not as optimal as proper dictionary reuse and delta support but will unlock ability to leverage dictionary batches over flight. Related work in Rust Arrow implementation: [https://github.com/apache/arrow-rs/pull/4896](https://github.com/apache/arrow-rs/pull/4896).

Copied from [https://github.com/apache/arrow/issues/38045](https://github.com/apache/arrow/issues/38045).

PR #433, opened 2026-09-10
feat: support reading dictionary-encoded columns over Arrow Flight

owencorrigan:
## What's Changed

The C# Arrow Flight reader (`RecordBatchReaderImplementation`) threw
`NotImplementedException` on any dictionary batch, so a Flight stream containing a
dictionary-encoded column could not be read.

- `ReadNextRecordBatchAsync` now loops over the incoming Flight messages, feeding
  each through the base `CreateArrowObjectFromMessage` (which reads dictionary
  batches into the `DictionaryMemo` and returns null for them) until it yields a
  record batch, instead of throwing on any non-record-batch message. Handles both
  single and replacement dictionaries.
- The schema is decoded using the reader's `DictionaryMemo` via a new base helper
  `ReadSchemaFromMessage`, so dictionary-encoded fields are registered
  (`FlightMessageSerializer` previously discarded them).

The change routes dictionary batches to the existing `ArrowReaderImplementation`
decoder, so it is not specific to any particular index or value type. Validated
against a pyarrow `RecordBatchStream` serving a `dictionary<int16, string>` column
(single + replacement dictionaries); the current reader throws on the same stream.

Refs #180 (receive side).

Generated-By: Claude Opus 4.8


CurtHagenlocher (changes requested):
Thanks! This would definitely benefit from some tests.

owencorrigan:
I've added support for cancellation. I have also added support for sending dictionary encoded columns over flight so that we could do a round trip test
Let me know if this is OK

CurtHagenlocher:
A round-trip test sounds like a good idea. Can you also please rebase and then resolve the conflicts?

